PR #1, opened 2021-02-04
start http and html skelleton

cryptix:
ground work for router and templating setup.

TODO:
* [x] tests (98ca4f8)
* [x] i18n
* [x] embedded assets (css, js, etc.)
* [x] sqlite
* [x] basic auth as a stub for _sign-in with ssb_


README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This repository contains code for a [Secure-Scuttlebutt Room v2](github.com/ssb-ngi-pointer/rooms2/) server writen in Go.
+
+It not only includes the secret-handshake+boxstream setup and muxrpc handlers for tunneling connections but also a fully embedded http/html interface for administering the room.
+
+## Features
+
+* [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
+* [ ] Sign-in with SSB
+* [x] Simple whitelisting
```
staltz, line 11: I know this was probably not an intentional use of the word whitelisting, but "allowlist"/"blocklist" are considered less racist words for whitelist/blacklist. :)

cryptix:
Nearly done. Will add _logged in?_ template helper tomorrow to show a log out button conditionally on the users state. Then it should be good to merge.

cryptix:
Okay. This should be it.

One missing step is creating the database on the first run. I will tackle this in a follow-up because this is already too big.

The workaround is basically `sqlite3 $HOME/.ssb-go-room/roomdb < $src/admindb/sqlite/schema-v1.sql`.

PR #16, opened 2021-02-10
web/i18n: Add defaults and test for labels (fixes #8)

### LINKED ISSUE #8: Bug: localization managment
Currently there are no defaults for the i18n system.

I propose to add the same mechanic that we use for migrations and web assets (to bundle them in the binary) but to also look into the repo folder (default `$HOME/.ssb-go-room`) for overrides so that users can more easily customize them.

cryptix:
@staltz I will leave it up to you how deep we want to go with this but this gives us `web/i18n/defaults` and we just test that the correct labels are used in the templating.


web/i18n/helper.go
```diff
@@ -39,10 +44,44 @@ func New(r repo.Interface) (*Helper, error) {
...
+		_, err = bundle.ParseMessageFileBytes(mfb, path)
 		if err != nil {
 			return fmt.Errorf("i18n: failed to parse file %s: %w", path, err)
 		}
+		fmt.Println("loaded", path)
+		return nil
+	}
+
+	// walk the embedded defaults
+	err := vfsutil.WalkFiles(Defaults, "/", walkFn)
```
staltz, line 60: Just trying to understand: where does `Defaults` come from?

staltz:
@cryptix Looks good to me. What's up with the tests in CI?

web/i18n/helper.go
cryptix: in `-tags dev` builds it's from here:

https://github.com/ssb-ngi-pointer/go-ssb-room/pull/16/files#diff-6eb611b44b73353aabbe7c47960117cf3e6dd36dfc697af131048df497eb7b44R23

otherwise it's from the defaults_vfsdata.go (updated with `go generate`)

cryptix:
Yup, that's odd.

The failing test is this one: https://github.com/ssb-ngi-pointer/go-ssb-room/blob/f0a785a800c2ca257cc469f86031ae4091df0f9d/internal/broadcasts/broadcast_test.go#L64 but this change doesn't touch it so I suspect this was already in master.

I was able to trigger it twice in `internal/broadcasts` using `go test -count 100000 -race` but then not anymore.

 Looking into it but merging this.

Issue #9, opened 2021-02-09
Integrate room details into web frontend


cryptix:
At a minimum it should show the ID of the room, and maybe it's `host:port` somewhere.

Maybe the count of people in the room (to be revisited once privacy settings are added).

cryptix:
Maybe a list of connected peers and their last ping on /admin/dashboard

PR #17, opened 2021-02-10
web room bridge

cryptix:
This should show a list of id that are in the room on the dashboard.


cryptix:
the big part of this was beea19f - there now is the roomstate package which implements a Manager, which supplies the following methods:

```go
func (m *Manager) AddEndpoint(who refs.FeedRef, edp muxrpc.Endpoint)
func (m *Manager) Has(who refs.FeedRef) (muxrpc.Endpoint, bool)
func (m *Manager) List() []string
func (m *Manager) Register(sink broadcasts.RoomChangeSink)
func (m *Manager) Remove(who refs.FeedRef)
```

Has(ref) can be used to get a muxrpc client to that peer, if it's connected. Right now only List() is used on the dashboard landing page.

cryptix:
Fixes #9

staltz (approved):

PR #19, opened 2021-02-11
setup Tailwind css project in web/styles

staltz:
@cryptix Most of this PR is in shape to merge (UPDATE: after rebasing, of course), the only thing I'd like to fix is removing the 3MB compiled.dev.css encoded directly inside `assets_vfsdata.go`, but I haven't figured out how to do that. `vfsgen` doesn't seem to allow me to customize what to include/exclude.


staltz:
Yeah these generated Go files have to leave Git, otherwise we'll do these git conflict resolutions in literally every PR.

cryptix:
I think it works right but can you try 9019ebd?

You can now use 'go generate -tags dev' to produce the dev build and 'go generate'
for the prod build. I changed it so that  they both produce the same file.

The only problem is that one needs to run this in web/styles first before updating assets.

staltz:
I'm out of "office" now, but conceptually this works for me. I haven't ran it though. Feel free to merge

staltz:
Oops, should have ran the tests and paid attention to the recent id changes.

staltz:
@cryptix I noticed that after [this commit](https://github.com/ssb-ngi-pointer/go-ssb-room/pull/19/commits/9019ebd130604834029a7f2ffe5f9a1eb868a652) my workflow to use `compiled.dev.css` changed. It took me a while to get used to the new way.

Going forwards, could we agree on doing the following pull request protocol? (we use this with arj)

1. PR author pushes to the branch and asks for feedback
2. Other person gives feedback and uses GH's 'approve' or 'request changes' UI
3. PR author pushes new commits to meet the feedback
4. Other person eventually 'approves' the PR
5. PR author merges

If the other person wishes to push commits to the PR, that's totally possible, but they should ask and/or inform the PR author.

Issue #7, opened 2021-02-09
Feature: allow-list managment


cryptix:
currently there is code that parses an authorized_keys file for public keys that are allowed to connect with the room.

This should be replaced with the admindb and some managment on the web frontend.

PR #22, opened 2021-02-15
Allow list

cryptix:
added `AllowList` interface to package admindb, and implemented in sqlite.

```go
// AllowListService changes the lists of people that are allowed to get into the room
type AllowListService interface {
	// Add adds the feed to the list.
	Add(context.Context, refs.FeedRef) error

	// HasFeed returns true if a feed is on the list.
	HasFeed(context.Context, refs.FeedRef) bool

	// HasFeed returns true if a feed is on the list.
	HasID(context.Context, int64) bool

	// List returns a list of all the feeds.
	List(context.Context) (ListEntries, error)

	// RemoveFeed removes the feed from the list.
	RemoveFeed(context.Context, refs.FeedRef) error

	// RemoveID removes the feed for the ID from the list.
	RemoveID(context.Context, int64) error
}
```

Also began with some lite adding, listing and removing in HTML.


cryptix:
Nearly done. I just have to solve one more fluke in the router.

cryptix:
I had to drop the named variables in routes.. See the commit message of 58f795d for the details.

cryptix:
This code already is hooked into the muxrpc server and replaces the `authorized_keys` file i whipped up first.

This fixes #7

cryptix:
@staltz if you have opinions or a rough guideline for what classes I should use for the templates here, now is your time :)

staltz:
@cryptix Looking good! I think for purposes of testing, just use HTML id, as you are using already it seems. We can use `class` only for styling. If we bump into the need of having multiple elements with the same ID, then we can look into that specific case, whether changing the "GoQuery" selectors to use tagnames or classes.

web/router/admin.go
```diff
@@ -15,8 +20,12 @@ func Admin(m *mux.Router) *mux.Router {
 		m = mux.NewRouter()
 	}
 
-	m.Path("/").Methods("GET").Name(AdminDashboard)
-	// m.Path("/settings").Methods("GET").Name(AdminSettings)
+	m.Path("/dashboard").Methods("GET").Name(AdminDashboard)
+
+	m.Path("/allow-list").Methods("GET").Name(AdminAllowListOverview)
```
staltz, line 25: The name allow-list is totally fine for implementation, it can stay like that. I was thinking that for user-facing situations, we would call this the "members" list. Like when someone gets added to a room, they become a room *member*. URLs are not the most visible "user interface" but it would make it easier for users to understand what the URL is if it were set to `/members`, `/members/add`, `/members/confirm`, `/members/remove`.

What do you think about changing these URLs here, while keeping the rest of the code using the name AllowList?

cryptix: > What do you think about changing these URLs here, while keeping the rest of the code using the name AllowList?

Yup, right. Agreed. Do we have a candidate for the denied/blocked ones?

cryptix:
Next steps: If we ware fine with this `AllowList` interface, I'd rename it to `AccessControlList` and just have multiple instances of it that are used accordingly in the authentication.

but before that we need to build a bit more configration tooling.

web/router/admin.go
staltz: Those we be called `banned` (and the verb, "to ban"), to distinguish banning from blocking. Banning is about rejecting connections (either IP-based or ID-based or both).

PR #25, opened 2021-02-17
CSRF and CSP

### LINKED ISSUE #14: Common web security practicies
* [x] CSRF via [gorilla/csrf](github.com/gorilla/csrf/)
* [x] Content Security Policy (CSP)
* [ ] Cross-Origin Resource Sharing (CORS)

cryptix:
Done:

* [x] CSRF
* [x]  TLS redirects and sanity checks
* [x] HSTS

fixes #14


web/handlers/http.go
```diff
@@ -152,11 +154,26 @@ func New(
 		auth.SetStore(store),
 		auth.SetErrorHandler(authErrH),
 		auth.SetNotAuthorizedHandler(notAuthorizedH),
+		auth.SetLifetime(2*time.Hour), // TODO: configure
```
cryptix, line 157: @staltz this was also why we were seeing these short session times.

open to better defaults.

staltz: Makes sense!

cryptix:
Should be good to merge but I want to check a couple of things (like the HSTS implications) once my brain is less foggy.

cryptix:
Okay.. I just checked and  realized that `default-src` means no inline styling.

cryptix:
Also need to fix mime-types on assets.

Issue #23, opened 2021-02-16
"flash" errors


cryptix:
not in the sense of _adobe flash_ but a way to add error/validation messages to a page (like name taken or login incorrect).

Right now we just render an error page with a back button.

cryptix:
Not sure if it's even needed at this stage but feels like a massive _quality of life_ improvement for users later.

staltz:
Yes, I thought about this too, like if you input an incorrect SSB ID. Not a big priority, but a nice to have.

PR #31, opened 2021-02-22
Prettify members page

staltz:
Will have to rebase this on `master` but I'll do that after dev mode is back (I need it for quickly developing with tailwind). So this PR is in draft mode.

Screenshot from 2021-02-22 17-47-47

### DIFF
--- a/web/handlers/admin/allow_list.go
+++ b/web/handlers/admin/allow_list.go
@@ -65,6 +65,10 @@ func (h allowListH) overview(rw http.ResponseWriter, req *http.Request) (interfa
 	if err != nil {
 		return nil, err
 	}
+	// Reverse the slice to provide recent-to-oldest results
+	for i, j := 0, len(lst)-1; i < j; i, j = i+1, j-1 {
+		lst[i], lst[j] = lst[j], lst[i]
+	}
 
 	return map[string]interface{}{
 		csrf.TemplateTag: csrf.TemplateField(req),
--- a/web/handlers/admin/allow_list_test.go
+++ b/web/handlers/admin/allow_list_test.go
@@ -31,7 +31,7 @@ func TestAllowListEmpty(t *testing.T) {
 	assertLocalized(t, html, []localizedElement{
 		{"#welcome", "AdminAllowListWelcome"},
 		{"title", "AdminAllowListTitle"},
-		{"#allowListCount", "ListCountPlural"},
+		{"#allowListCount", "MemberCountPlural"},
 	})
 }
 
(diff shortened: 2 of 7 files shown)


web/handlers/admin/allow_list.go
```diff
@@ -64,6 +64,10 @@ func (h allowListH) overview(rw http.ResponseWriter, req *http.Request) (interfa
 	if err != nil {
 		return nil, err
 	}
+	// Reverse the slice to provide recent-to-oldest results
+	for i, j := 0, len(lst)-1; i < j; i, j = i+1, j-1 {
+		lst[i], lst[j] = lst[j], lst[i]
+	}
```
staltz, line 70: @cryptix What do you think about this?

cryptix: yup, nice use of the multi assign.

If you would want to sort by something other then the index of the array, I'd say go for [sort.Reverse](https://golang.org/pkg/sort/#Reverse) but this is fine until we change `List(ctx)` to have options and let the database do the work.

staltz: Oh, didn't know about `sort.Reverse`. Yeah, and I'll revisit this code once I work on pagination

cryptix: you would have to implement sort.Interface, like [here](https://github.com/cryptoscope/ssb/blob/f8b517c21b5d9b8fb05f923f29e59dda9a1730e5/sbot/status.go#L66).

cryptix (approved):

staltz:
Thanks for reviewing! I'll rebase tomorrow. Signing-off from work for today :)

staltz:
@cryptix This is ready, I added one commit to format the HTML. Updating to Go 1.16 was pretty straightforward.

One thing I noticed is that links such as

```html
<a href="javascript:history.back()" class="btn btn-primary">Back</a>
```

stopped working after the (I guess) CSP PR. Do you know how/why?

staltz:
And... I'll work on fixing the tests. I should have ran them locally.

cryptix:
> Do you know how/why?

Yup, you guessed right. That is inline code which is outlawed with the default CSP we adopted.

ps: i can shed some light on alternatives: If you really want inline content wit CSP, the options boil down to _nonces_ (which you need to add to the tag and set the nonce in the http header) or _hash the content of the tags_, which both requires quite the complex templating system setup to pull of. The third option is _unsafe_ which disables the protection all together.

cryptix:
PS: on the tests, see what makes sense.

I just tried to show some examples of how we can test the content of the pages with goquery.

cryptix:
on the back button specifically, I'd say we want to adopt #23 together with putting the _back redirect url_ in the template.

Then we don't have to involve page JS at all.

staltz:
> on the back button specifically, I'd say we want to fix #23 together with putting the _back redirect url_ in the template.
> 
> Then we don't have to involve page JS at all.

I'm happy with this option, and I'm okay if the back button is broken in the meanwhile. There's anyway the browser back button.

PS: fixed the tests

staltz:
Also, the HTML format is a bit strange, but it's a fine balance:

```html
<span
  id="welcome"
  class="text-center"
>{{i18n "AdminAllowListRemoveConfirmWelcome"}}</span>
```

There can't be spaces and newlines in the "content" portion because it would affect tests, and sometimes affects even layout.

cryptix:
> There's anyway the browser back button.

yup. It's mostly a QOL thing, subjecting people to the obnoxious _do the POST again?!..._ warning.

staltz:
Let's merge this?

cryptix:
Sorry, was half way out for launch already.. :sweat_smile:

PR #36, opened 2021-02-23
highlight active menu item with current_page_is

staltz:


web/handlers/admin/app_test.go
```diff
@@ -53,6 +53,9 @@ func newSession(t *testing.T) *testSession {
 		}
 		return msgID + "Plural"
 	}
+	testFuncs["current_page_is"] = func(routeName string) bool {
+		return true
```
staltz, line 57: @cryptix Help? I don't know how to get the "test request" here.

cryptix: Hmmmm. lemme se.

This might be fine to just return true or false if we decide to test this _just_ in `web/handlers` not each and every sub-system.

cryptix: That is basically also why only the _complete_ page in `web/handlers/http_test.go` has a real `is_logged_in` and the others have stubs.

The admin tests all assume a loggedin session. The one above tests the session behavior.

staltz: Yeah, for this case we could just return true, but maybe one day we have critical info displayed on the page depending on `current_page_is`. Just checking what your thoughts were about retrieving the request info

cryptix: > maybe one day we have

In full transparency, my long range predictions are a bit hit and miss in that regard.. :sweat_smile: 

These days I prefer to keep it simple and to re-evaluate once we get to those points where the use-cases become bigger and concrete.

staltz: Yep, I'm fine with the `return true` if you're fine

cryptix: I am :)

Issue #34, opened 2021-02-23
When to collapse the menu layout


cryptix:
I'm not a super strong proponent of this but I sympathize with the gist of this article:

[Just because I have a vertical screen, doesn’t mean I’m on a phone!](https://shkspr.mobi/blog/2021/02/just-because-i-have-a-vertical-screen-doesnt-mean-im-on-a-phone/)

(I use slim column layouts on my wide screen a lot)

staltz:
I don't understand, and I have a vertical screen. Can you clarify?

cryptix:
Oh.. I thought I saw it differently on the screenshot you posted on #32.

I thought the list of menu items would switch to horizontal on slim layouts.

I have a hunch there will be more buttons and if it stays vertical there will be a lot of white-space to the right of the menu items.

cryptix:
Also, I feel like it could switch to collapsed sooner but that might be preference.

staltz:
So this issue is that on mobile (and slim) screens there should be menu items side-by-side, not vertically?

staltz:
I guess this is closed?

PR #38, opened 2021-02-23
new route "menu" that is accessible on mobile only

staltz:
I found a neat solution to #34:

- When the layout is mobile-or-narrow, don't show the menu at all, instead show a menu icon on the top-left corner
- When the layout is wide, show the menu items and don't show the menu icon
- The menu icon takes you to `/admin/menu` which contains only the menu items

### Desktop looks like

Screenshot from 2021-02-23 20-20-31

### Mobile looks like

Screen Shot 2021-02-23 at 20 20 24

### `/admin/menu` looks like

Screen Shot 2021-02-23 at 20 20 40

### DIFF
--- a/web/handlers/admin/handler.go
+++ b/web/handlers/admin/handler.go
@@ -16,6 +16,7 @@ import (
 
 var HTMLTemplates = []string{
 	"admin/dashboard.tmpl",
+	"admin/menu.tmpl",
 	"admin/allow-list.tmpl",
 	"admin/allow-list-remove-confirm.tmpl",
 }
@@ -31,6 +32,9 @@ func Handler(r *render.Renderer, roomState *roomstate.Manager, al admindb.AllowL
 			Clients []string
 			Count   int
 		}{lst, len(lst)}, nil
+	}))
+	mux.HandleFunc("/menu", r.HTML("admin/menu.tmpl", func(w http.ResponseWriter, req *http.Request) (interface{}, error) {
+		return map[string]interface{}{}, nil
 	}))
 
 	var ah = allowListH{
--- a/web/handlers/admin/app_test.go
+++ b/web/handlers/admin/app_test.go
@@ -60,7 +60,7 @@ func newSession(t *testing.T) *testSession {
 
 	r, err := render.New(web.Templates,
 		render.SetLogger(log),
-		render.BaseTemplates("base.tmpl"),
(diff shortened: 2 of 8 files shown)


cryptix (commented):
Not sure we need the `/admin/menu` route.

I'm checking this out and will have a stab.

web/handlers/admin/handler.go
```diff
@@ -32,6 +33,9 @@ func Handler(r *render.Renderer, roomState *roomstate.Manager, al admindb.AllowL
 			Count   int
 		}{lst, len(lst)}, nil
 	}))
+	mux.HandleFunc("/menu", r.HTML("admin/menu.tmpl", func(w http.ResponseWriter, req *http.Request) (interface{}, error) {
```
cryptix, line 36: Technically this shouldn't be here. This should only make sense if we were using iframes to embed the template in the page and it should be possible without. The block/template system should be able to do this.

cryptix: Maybe i'm missing something and i'ts an intended feature to _just_ have the menu?

staltz: Yeah the intention is to have a screen or page that only shows the menu, because you asked for it via the menu button.

We can do this in JS and then we won't need the new route /admin/menu but we want to do things without JS right?

cryptix: oooooh, sorry. I was just wondering if we can lay it out as a list with items next to each other compared to over and under. I didn't think of a separate page per se but that could work, too.

I know people that would be glad with a JS-less page but personally I'm fine with..what's the term.. progressive enhancement? (it should still work without)

staltz: >  I was just wondering if we can lay it out as a list with items next to each other compared to over and under.

That would work now, but as soon as we add more items to the list (edit notices, manage bans, manage room settings) then it won't fit side-by-side on the top anymore. Also a menu button is quite common in these cases

staltz:
@cryptix Tests pass now, and git conflict resolved too.

web/handlers/admin/handler.go
staltz: My wish with not having JS is just to keep the codebase simple, easier to maintain and reason about when modifying the UI.

staltz:
@cryptix Is this PR approved?

cryptix (approved):
SGTM

PR #39, opened 2021-02-24
implement pagination in allow-list overview

staltz:
<details>
<summary>Click to see animated GIF</summary>

### Using pageSize=2 just for illustrative purposes

pagination

</details>


web/handlers/admin/allow_list.go
```diff
@@ -69,11 +74,53 @@ func (h allowListH) overview(rw http.ResponseWriter, req *http.Request) (interfa
...
+	if page < 1 {
+		page = 1
+	}
+
+	paginator := paginator.New(adapter.NewSliceAdapter(lst), pageSize)
+	paginator.SetPage(page)
+
+	var entries admindb.ListEntries
+	if err = paginator.Results(&entries); err != nil {
+		panic(err)
```
cryptix, line 93: just do `return nil, err` here instead of panic()s

if you are really fancy, the way to add context would be this:

```
return nil, fmt.Errorf("paginator failed with %w", err)`
```

where fmt.Errorf is just a fancy printf, so you could add all kinds of info to the error sting.

cryptix (approved):
Yup, LGTM.

The paginator also has an interface we could implement directly in the database when the need arises.

web/handlers/admin/allow_list.go
staltz: Ok, I didn't know what to do here, so I used a panic().

What would return `nil` do to the HTML renderer?

staltz: Pushed a commit with the new errors

cryptix: > What would return nil do to the HTML renderer?

If you also return an error it doesn nothing and renders the error.tmpl instead.

cryptix: The mentioned code path is here: https://github.com/cryptix/go/blob/eb625e0824c4572ba1c7e10251c61ba1e34731b6/http/render/templating.go#L109

I need to make a patch for `http/render` so that we can override the `r.Error` handler there. Then we can check for the error types and return localized error messages for well known stuff like ErrBadRequest.

cryptix (approved):
Sweet! LGTM

PR #54, opened 2021-03-02
fix CSS for long code blocks

staltz:
## Before

Screenshot from 2021-03-02 16-27-00

## After

Screenshot from 2021-03-02 16-29-01

### DIFF
--- a/web/styles/input.css
+++ b/web/styles/input.css
@@ -63,3 +63,7 @@ html {
 .markdown hr {
   @apply my-5;
 }
+
+.markdown code {
+  @apply break-all;
+}


cryptix (approved):
LGTM!

Just saw the last part, I _think_ we can punt on syntax highlighting. There is support for it in blackfriday using [chroma](https://github.com/Depado/bfchroma/) but I don't see a real need for it.

staltz:
> Just saw the last part, I _think_ we can punt on syntax highlighting. There is support for it in blackfriday using [chroma](https://github.com/Depado/bfchroma/) but I don't see a real need for it.

Oh totally, that was in my example just because I copy pasted from somewhere.

Issue #13, opened 2021-02-10
Invite creation CRUD in the dashboard


staltz:

cryptix:
I will start with the [community mode of joining](https://github.com/ssb-ngi-pointer/rooms2/blob/4915c544a5bf902aff0dc449151424591180ec04/docs/Participation/Joining.md#L1) since we don't have the different modes yet. Restricting form there based on user roles later seems to make the most sense for me.

* [ ] add _create invite_ button
  * [ ] HTTP endpoint
  * [ ] outline `admindb.InviteService` and SQLite schema
  * [ ] dashboard link
  * [ ] _here is your invite_ HTML template
  * [ ] also as JSON for better App integration
* [ ] add _consume_ endpoint
  * [ ] HTTP endpoint
  * [ ] update active invite token (possible with the public key of the consumee so that the invite creator can follow/allow them)
  * [ ] returns the [tunnel address](https://github.com/ssb-ngi-pointer/rooms2/blob/4915c544a5bf902aff0dc449151424591180ec04/docs/Participation/Tunnel%20addresses.md)
  * [ ] either as a HTML page or ssb-uri redirect depending on the request

staltz:
Is this done? It seems done

cryptix:
Indeed!

PR #53, opened 2021-03-02
disambiguate "handlers" package

cryptix:
This was fine when there was just one for muxrpc but "handlers"
and "web/handlers" feels like asking for confusion.

```
[clarify-handlers 6132b37] disambiguate "handlers" package
 Date: Tue Mar 2 14:17:30 2021 +0100
 6 files changed, 6 insertions(+), 2 deletions(-)
 create mode 100644 muxrpc/doc.go
 rename {handlers => muxrpc}/tunnel/server/connect.go (100%)
 rename {handlers => muxrpc}/tunnel/server/plugin.go (100%)
 rename {handlers => muxrpc}/tunnel/server/state.go (100%)
 rename {handlers => muxrpc}/whoami/whoami.go (100%)
```

PR #55, opened 2021-03-02
admindb: InviteService

cryptix:
updates #13 

* [x] enable foreign key enforcement on sqlite
* [x] admindb.InviteService interface
  * [x] Create
  * [x] Consume
  * [x] List
  * [x] Revoke
* [x] tests for the sqlite implementation of the service


cryptix:
As these things go, most of the diff is from the new generated code for invites in mockdb and sqlite/models (sqlboiler).

admindb/sqlite/new.go
```diff
@@ -1,5 +1,19 @@
 // SPDX-License-Identifier: MIT
 
+// Package sqlite implements the SQLite backend of the admindb interfaces.
+//
+// It uses sql-migrate (github.com/rubenv/sql-migrate) for it's schema definition and maintainace.
+// For query construction/ORM it uses SQLBoiler (https://github.com/volatiletech/sqlboiler).
+//
+// The process of updating the schema and ORM can be summarized as follows:
```
cryptix, line 8: I think I never spelled out how updating the sqlite impl was supposed to go. Hope this makes it clearer.

cryptix:
I'd prefer if we use [godoc](https://blog.golang.org/godoc) for package and function level documentation and READMEs for meta/setup level stuff.

I usually use `go doc admindb` for instance or `go doc admindb AllowListServe` from anywhere in the repo. If package names collide, you can use the full import path, like  (at which point I usually prefer browsing `https://pkg.go.dev/import/path/..`, unless it's in the stdlib, since it's a a lot to type and browsers have a better history and completion.

admindb/sqlite/auth_fallback.go
```diff
@@ -39,23 +39,23 @@ func (ah AuthFallback) Check(name, password string) (interface{}, error) {
 	return found.ID, nil
 }
 
-func (ah AuthFallback) Create(ctx context.Context, name string, password []byte) error {
+func (ah AuthFallback) Create(ctx context.Context, name string, password []byte) (int64, error) {
```
cblgh, line 42: I wonder if we can have a type for the invite, instead of obliquely passing around `int64`–or is that a bit much?

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/volatiletech/null/v8"
+	"github.com/volatiletech/sqlboiler/v4/boil"
+	"github.com/volatiletech/sqlboiler/v4/queries/qm"
+
+	"github.com/ssb-ngi-pointer/go-ssb-room/admindb"
+	"github.com/ssb-ngi-pointer/go-ssb-room/admindb/sqlite/models"
+	refs "go.mindeco.de/ssb-refs"
+)
+
+// make sure to implement interfaces correctly
```
cblgh, line 22: could this scary comment be clarified within a parens? 
>  // make sure to implement interfaces correctly (e.g. don't use sqlboiler generated types in args/return values)

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+var _ admindb.InviteService = (*Invites)(nil)
+
+// Invites implements the admindb.InviteService.
+// Tokens are stored hashed with sha256 on disk to protect against someone who gaining read-access on the database.
```
cblgh, line 26:
```suggestion
// Tokens are stored as sha256 hashes on disk to protect against attackers gaining database read-access.
```

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+type Invites struct {
+	db *sql.DB
+
+	allowList *AllowList
+}
+
+const tokenLength = 50
```
cblgh, line 33: magic number? or why is it 50?

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Create creates a new invite for a new member. It returns the token or an error.
+// createdBy is user ID of the admin or moderator who created it.
+// aliasSuggestion is optional (empty string is fine) but can be used to disambiguate open invites. (See https://github.com/ssb-ngi-pointer/rooms2/issues/21)
+// The returned token is base64 URL encoded and has tokenLength when decoded.
+func (i Invites) Create(ctx context.Context, createdBy int64, aliasSuggestion string) (string, error) {
+	isValidSuggestion := aliasSuggestion != ""
+	var newInvite = models.Invite{
+		CreatedBy:       createdBy,
+		AliasSuggestion: null.NewString(aliasSuggestion, isValidSuggestion),
```
cblgh, line 43: `null.NewString` *does a double take*, [oh](https://github.com/volatiletech/null)

admindb/sqlite/auth_fallback.go
cryptix: Right. Technically this leaks (or requires) that all the admindbs have numerical ID types.

I honestly don't think it makes much sense to put another level of indirection here. (The way out would be to have interfaces but what would the method on that be but returning an integer again or do a bunch of assertions to get the actual value back from underlying implementations)

We could have `type ID int64` in admindb to make it more concrete what these numbers are (could also be affected rows or something entirely different) they will need to be cast like `int64(userID)` in more cases then one would think. Go is quite picky there.

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	tokenBytes := make([]byte, tokenLength)
+
+	err := transact(i.db, func(tx *sql.Tx) error {
+
+		inserted := false
+	trying:
+		for tries := 100; tries > 0; tries-- {
+
```
cblgh, line 53:
```suggestion
```

admindb/sqlite/invites.go
cryptix: Ah. These two lines are a compiler guarantee thing.  It just checks that the struct fulfills the interface. (You assign to an unused variable with a nil thing, which doesn't even allocate anything.)

meaning if you change the methods slightly this will fail to compile and tell you _I expected method X to return Y, not Z_.

Still open for suggestions how to phrase that better.

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,207 @@
...
+func (i Invites) Consume(ctx context.Context, token string, newMember refs.FeedRef) (admindb.Invite, error) {
+	var inv admindb.Invite
+
+	tokenBytes, err := base64.URLEncoding.DecodeString(token)
+	if err != nil {
+		return inv, err
+	}
+
+	if n := len(tokenBytes); n != tokenLength {
+		return inv, fmt.Errorf("admindb: invalid invite token length (only got %dbytes)", n)
```
cblgh, line 102:
```suggestion
 		return inv, fmt.Errorf("admindb: invalid invite token length (only got %d bytes)", n)
```

admindb/sqlite/invites.go
cryptix: The rooms2 doc leaves this open. cc @staltz

admindb/sqlite/invites.go
cryptix: We could possible just store empty strings. They don't _need_ to be stored as NULLs when not present.

admindb/sqlite/invites.go
staltz: What's the size of normal pub invite codes? Should be the same as this one too, no? Is this something that needs to be spec'd? (I don't see why all the rooms would have to have the same lengths)

cryptix: Same for the choice of encoding the tokens. url safe base64 seemed like a better choice then hex.

Also not super sure it needs to be defined strictly since it just needs to be returned and passed back as an opaque string.

cryptix: > What's the size of normal pub invite codes? 

In classic ssb-invites it's a seed for a keypair for ed22519, which is 32bytes. But that protocol is not susceptible to brute forcing an HTTP endpoint, which is why I made it longer.

admindb/sqlite/auth_fallback.go
cblgh: > We could have type ID int64 in admindb to make it more concrete what these numbers are (could also be affected rows or something entirely different) they will need to be cast like int64(userID) in more cases then one would think. Go is quite picky there.

Yeah that was basically the only thing I was advocating for, just to make the interfaces a bit more readable & self-documenting, but if you think it would cause a lot of casts then I don't think it is worth it here.

*resolved*

admindb/interface.go
```diff
@@ -1,5 +1,13 @@
 // SPDX-License-Identifier: MIT
 
+// Package admindb implements all the persisted database needs of the room server.
```
cblgh, line 3: One thing I kept thinking: since this package takes care of persisting all the data, then maybe `admindb` isn't really descriptive? `roomdb` might be better; `admindb` implies being concerned with a database of admins, I think, suggesting the presence of other databases as well. 

I didn't really want to comment on this though since it seems so nitpicky, but I kept thinking about it so 🤷

cryptix: :+1: 

I had a similar thought when I wrote this but didn't want to overboard this PR. I will do it in a follow up similar to #53

admindb/sqlite/invites.go
cblgh: ah ok yeah then maybe persisting what you wrote above as a comment could make sense? e.g.
```suggestion
// Checks that the struct fulfills the interface. (You assign to an unused variable with a nil thing, which doesn't even allocate anything.)
```

cause i was just spooked as hell by the comment, especially when my lack of internal project context was combined with the top-most package comment re sqlboiler generated variables

admindb/interface.go
cblgh: cool, glad to hear it validated something you were thinking of! :) i was really hesitant to comment on it haha

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,206 @@
...
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(tokenBytes), nil
+}
+
+// Consume checks if the passed token is still valid. If it is it adds newMember to the members of the room and invalidates the token.
+// If the token isn't valid, it returns an error.
+// Tokens need to be base64 URL encoded and when decoded be of tokenLength.
+func (i Invites) Consume(ctx context.Context, token string, newMember refs.FeedRef) (admindb.Invite, error) {
```
cblgh, line 90: is it super faux-pas to do like `func (this Invites) ... <function>`? `i` is so ingrained for me as an iteration variable / useless int haha, had to do a double-take when reading the code & skimming the function signature. kinda on the nitpicky side tho

(or `func (invites Invites)` but borders on verbose)

cblgh: also i recently read https://macwright.com/2021/02/17/the-naming-of-things.html so i think that's definitely a lens i am reviewing through ^^

cblgh (commented):
👍 👍 👍

admindb/sqlite/invites.go
cryptix: hm yea.. i don't know why but i really don't like `this` or `self` for these. Definitely a style preference thing.

I was starting out with `db` but since in most cases here the only thing in the receiver struct is the database handle `db`, I ended up with a lot of `db.db` which also felt silly.

admindb/interface.go
cryptix: I created #57, too :)

admindb/sqlite/auth_fallback.go
cryptix: I think it should be fine until we end up with another class/group of integers we have to work with and so far I don't see that happening in this project.

admindb/sqlite/invites.go
```diff
@@ -0,0 +1,206 @@
...
+		entries, err := models.Invites(
+			qm.Where("active = true"),
+			qm.Load("CreatedByAuthFallback"),
+		).All(ctx, tx)
+		if err != nil {
+			return err
+		}
+
+		invs = make([]admindb.Invite, len(entries))
+		for i, e := range entries {
```
cblgh, line 159: overloading `i Invites` with `i` range count

admindb/sqlite/invites_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/ssb-ngi-pointer/go-ssb-room/internal/repo"
+	"github.com/stretchr/testify/require"
+	refs "go.mindeco.de/ssb-refs"
+)
+
+func TestInvites(t *testing.T) {
+
```
cblgh, line 18:
```suggestion
```

admindb/sqlite/invites_test.go
```diff
@@ -0,0 +1,111 @@
...
+	ctx := context.Background()
+
+	testRepo := filepath.Join("testrun", t.Name())
+	os.RemoveAll(testRepo)
+	tr := repo.New(testRepo)
+
+	// fake feed for testing, looks ok at least
+	newMember := refs.FeedRef{ID: bytes.Repeat([]byte("acab"), 8), Algo: refs.RefAlgoFeedSSB1}
```
cblgh, line 26: `bytes.Repeat([]byte("acab")`

🤣

admindb/sqlite/invites.go
cryptix: yikes. yea, that's bad. 

I just checked and noticed the other sqlite sub-service also have inconsistent names.

How about having `s` for service as the last letter?

So
* `func (is Invites) ...`
* `func (als AllowList) ...`
* `func (afs AuthFallback) ..`
* `func (as AliasService) ...`
* `func (ns Notes) ...`


Similar to the package name change I'd prefer this as a follow up, though.

cblgh: oops didn't respond: i really like the idea of keeping things cohesive! i'm ambivalent on using `s`, i don't really know the code well enough yet to have a feeling for how it works. some weird semantic clashes with `is`, `as`, `ns` but :shrug:

admindb/sqlite/migrations/4-invites.sql
```diff
@@ -0,0 +1,23 @@
+-- +migrate Up
+CREATE TABLE invites (
+  id               integer PRIMARY KEY AUTOINCREMENT NOT NULL,
+  token            text UNIQUE NOT NULL,
+  created_by       integer NOT NULL,
+  alias_suggestion text NOT NULL DEFAULT "", -- optional
+
+  active boolean not null default true,
```
cblgh, line 8:
```suggestion
  active boolean NOT NULL DEFAULT true,
 ```

File: admindb/sqlite/migrations/4-invites.sql [hunk 15]
```diff
@@ -0,0 +1,23 @@
...
+
+  -- TODO: replace auth_fallback with a user table once we do "sign in with ssb"
+  FOREIGN KEY ( created_by ) REFERENCES auth_fallback( "id" )
+);
+
+CREATE INDEX invite_active_ids ON invites(id) where active=true;
```
cblgh, line 14:
```suggestion
CREATE INDEX invite_active_ids ON invites(id) WHERE active=true;
 ````

File: admindb/sqlite/new.go [hunk 16]
```diff
@@ -57,14 +80,41 @@ func Open(r repo.Interface) (*Database, error) {
...
+			err := transact(db, func(tx *sql.Tx) error {
+				_, err := models.Invites(qm.Where("active = false")).DeleteAll(context.Background(), tx)
+				return err
+			})
+			if err != nil {
+				// TODO: hook up logging
+				log.Printf("admindb: failed to clean up old invites: %s", err.Error())
+			}
+		}
+	}()
```
cblgh, line 102: wut

cblgh (approved):
💯  batch the suggested changes you agree with, discard the rest & 📦 🚢 it!!

PR #77, opened 2021-03-11
Flesh out development instructions

cblgh:
Added some instructions on how to get started viewing things locally, and rearranged content to make it more easily perceptible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,28 +1,32 @@
 # Go-SSB Room
 
-This repository contains code for a [Secure-Scuttlebutt Room v2](https://github.com/ssb-ngi-pointer/rooms2/) server written in Go.
+This repository contains code for a [Secure Scuttlebutt](https://ssb.nz) [Room (v1+v2) server](https://github.com/ssb-ngi-pointer/rooms2), written in Go.
 
-It not only includes the secret-handshake+boxstream setup and muxrpc handlers for tunneling connections but also a fully embedded http/html interface for administering the room.
+It includes:
+* a secret-handshake+boxstream setup
+* muxrpc handlers for tunneling connections
+* a fully embedded server & frontend, for administering the room
 
 ## Features
 
 * [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
+* [x] Simple allow-listing   
+    Currently via `.ssb-go-rooms/authorized_keys`.   
+    To be replaced with an authorization list on the dashboard.
 * [ ] Sign-in with SSB
-* [x] Simple allow-listing
-    Currently via `.ssb-go-rooms/authorized_keys`.
-    To be replaced with a authorization list on the dashboard.
 * [ ] Alias managment
 
 ## Development
 
 To get started, you need a recent version of [Go](https://golang.org). v1.16 and onward should be sufficient.
 
-To build the server and see a list of it's options, run the following:
+To build the server and see a list of its options:
 
 ```bash
-$ cd cmd/server
-$ go build
-$ ./server -h
+cd cmd/server
+go build
+./server -h
+ 
 Usage of ./server:
   -dbg string
     	listen addr for metrics and pprof HTTP server (default "localhost:6078")
@@ -44,6 +48,9 @@ Usage of ./server:
 
 If you want to view the development server in your browser:
 ```sh
+# change to the root of the project (e.g. cd go-ssb-room) and generate the frontend's styling; requires npm
+go generate -tags dev ./...
+# now let's build & run the development server
 cd cmd/server && go build -tags dev && ./server
 # and visit http://localhost:3000
 ```
@@ -53,13 +60,13 @@ This can be useful if you are working on:
 * html templates, 
 * or website assets
 
-This way the build won't use the assets embedded in the binary, but instead read them directly from the local filesystem.
+This way, the build won't use the assets embedded in the binary, but instead read them directly from the local filesystem.
 
-Once you are done with your changes run and want to update the embedded assets:
+Once you are done with your changes and want to update the embedded assets:
 ```sh
-go generate
+# cd to the root of the folder, and then run go generate
+go generate -tags dev ./...
 ```
-**Note**: you need to run generate in each changed package.
 
 ## Tooling
 ### Mocks
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ It includes:
 * [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
 * [x] Simple allow-listing   
     Currently via `.ssb-go-rooms/authorized_keys`.   
-    To be replaced with an authorization list on the dashboard.
 * [ ] Sign-in with SSB
 * [ ] Alias managment
 
--- a/README.md
+++ b/README.md
@@ -3,15 +3,14 @@
 This repository contains code for a [Secure Scuttlebutt](https://ssb.nz) [Room (v1+v2) server](https://github.com/ssb-ngi-pointer/rooms2), written in Go.
 
 It includes:
-* a secret-handshake+boxstream setup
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,10 +42,24 @@ Usage of ./server:
...
+* or website assets
 
-Once you are done with your changes run `go generate` in the changed packages to update them.
+This way the build won't use the assets embedded in the binary, but instead read them directly from the local filesystem.
+
+Once you are done with your changes run and want to update the embedded assets:
+```sh
+go generate
+```
+**Note**: you need to run generate in each changed package.
```
cryptix, line 62: this can be `go generate ./...` from the root folder of the repo.

the only thing that is a big quirky is #52 - for development you need `go generate -tags dev` to get the correct tailwind css.

README.md
```diff
@@ -42,10 +42,24 @@ Usage of ./server:
...
+# and visit http://localhost:3000
+```
+
+This can be useful if you are working on:
+* the sqlite migrations, 
+* html templates, 
+* or website assets
 
-Once you are done with your changes run `go generate` in the changed packages to update them.
+This way the build won't use the assets embedded in the binary, but instead read them directly from the local filesystem.
```
cryptix, line 56: We changed to asset embedding via `embed.FS` using Go 1.16 (#26) so `go generate` is not necessary to update the embedded assets anymore, however there is #52 to solve still.

So the only important difference is building with `-tags dev` for development.

cryptix (commented):
Yup +1.

Just the confusion about the asset embedding that should have been removed after #26 already... :see_no_evil:

cblgh:
does the above look okay @cryptix? simplified things a bit more and rewrote it based on what we talked about yesterday (+ yr comments above)

cblgh:
how can my readme fail the pipeline

cryptix:
> how can my readme fail the pipeline

ugh... you are totally, right.

There are two _semi weird_ glitches with the network and logging code that only show itself on CI (I suspect because of high load of the boxes). If this happens, feel free to re-run them.

README.md
```diff
@@ -1,28 +1,32 @@
 # Go-SSB Room
 
-This repository contains code for a [Secure-Scuttlebutt Room v2](https://github.com/ssb-ngi-pointer/rooms2/) server written in Go.
+This repository contains code for a [Secure Scuttlebutt](https://ssb.nz) [Room (v1+v2) server](https://github.com/ssb-ngi-pointer/rooms2), written in Go.
 
-It not only includes the secret-handshake+boxstream setup and muxrpc handlers for tunneling connections but also a fully embedded http/html interface for administering the room.
+It includes:
+* a secret-handshake+boxstream setup
```
cryptix, line 6:
```suggestion
* secret-handshake+boxstream network transport, sometimes referred to as SHS, using [secretstream](https://github.com/cryptoscope/secretstream)
```

README.md
```diff
@@ -1,28 +1,32 @@
...
+* muxrpc handlers for tunneling connections
+* a fully embedded server & frontend, for administering the room
```
cryptix, line 8:
```suggestion
* a fully embedded HTTP server & HTML frontend, for administering the room
```

README.md
```diff
@@ -42,10 +46,27 @@ Usage of ./server:
     	print version number and build date
 ```
 
-If you are working on the sqlite migrations, html templates or website assets, build the server with `go build -tags dev`.
-This way it won't use the assets that are embedded in the binary but read them directly from the local filesystem.
+If you want to view the development server in your browser:
+```sh
+# change to the root of the project (e.g. cd go-ssb-room) and generate the frontend's styling; requires npm
+go generate -tags dev ./...
```
cryptix, line 52:
```suggestion
cd web/styles && go generate -tags dev
```

README.md
```diff
@@ -42,10 +46,27 @@ Usage of ./server:
...
+
+This can be useful if you are working on:
+* the sqlite migrations, 
+* html templates, 
+* or website assets
 
-Once you are done with your changes run `go generate` in the changed packages to update them.
+This way, the build won't use the assets embedded in the binary, but instead read them directly from the local filesystem.
+
+Once you are done with your changes and want to update the embedded assets:
```
cryptix, line 65: This updating is no longer required since the embedding happens at build time via the Go1.16 [embed](https://golang.org/pkg/embed/) package.

The previous setup generated `_vfsgen.go` files, which included the contents of the assets/templates folder but they are all gone now, which means, no updating required :)

README.md
```diff
@@ -1,28 +1,32 @@
...
+It includes:
+* a secret-handshake+boxstream setup
+* muxrpc handlers for tunneling connections
+* a fully embedded server & frontend, for administering the room
 
 ## Features
 
 * [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
+* [x] Simple allow-listing   
+    Currently via `.ssb-go-rooms/authorized_keys`.
```
cryptix, line 14:
```suggestion
```

Done and removed as of #22

README.md
```diff
@@ -1,28 +1,32 @@
...
-It not only includes the secret-handshake+boxstream setup and muxrpc handlers for tunneling connections but also a fully embedded http/html interface for administering the room.
+It includes:
+* a secret-handshake+boxstream setup
+* muxrpc handlers for tunneling connections
+* a fully embedded server & frontend, for administering the room
 
 ## Features
 
 * [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
+* [x] Simple allow-listing
```
cryptix, line 13:
```suggestion
* [x] Simple allow-listing, administerd via the web dashboard
```

README.md
```diff
@@ -1,28 +1,32 @@
...
+* a secret-handshake+boxstream setup
+* muxrpc handlers for tunneling connections
+* a fully embedded server & frontend, for administering the room
 
 ## Features
 
 * [x] Rooms v1 (`tunnel.connect`, `tunnel.endpoints`, etc.)
+* [x] Simple allow-listing   
+    Currently via `.ssb-go-rooms/authorized_keys`.   
+    To be replaced with an authorization list on the dashboard.
```
cryptix, line 15:
```suggestion
```

(See above)

cryptix (commented):
Thanks a ton! would be good to go with the suggestions merged.

README.md
cblgh: the original would work as well right? or would that cause extra work? not sure if we have generate statements elsewhere.

if the original would work i think it's better, just cause it's like "whatever just run the thing where it's needed" and you stay in the same directory

README.md
cblgh: in general in the intro: i was trying to keep it kinda minimal & tidy as it's the first thing people see :) guessing the majority of people won't care about the details, so much as having a nice room to deploy.

if the details are important for you in this part though, i'll just add them :)

README.md
cryptix: I think it could be debated if we want `go generate` here and not just document whats going on, which is:

```bash
cd web/styles
npm ci
npm run compile-dev
```

This why we remove the confusing `go generate` indirection that's not required for anything else anymore.

ps: people also need to know that they need to run this before building the server.


```bash
cd web/styles
npm ci
npm run compile-prod
```
(See #52 for some discussion about this)

README.md
cryptix: I think we just spell out what kind of frontend is to be expected and disambiguated that there is a muxrpc and a http server here but I don't think it's necessary.

README.md
cblgh: yup up to you, i think we should merge this and you can change it to what you want, separately :)

i never really liked the concept of debate class ;)

cryptix: Fine by me :) It definitely works the way you wrote it.

README.md
cblgh: ya i got that previously, but still a bit confused by the tailwind situation. so they are not being embedded? or are they?

cblgh:
:package:

Issue #68, opened 2021-03-10
Invites follow-up: page styling


cryptix:
The new pages to be styled after #59 are:

These in `web/handlers/admin`:

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/04824eeed78fedfd10b47e4e72081d18e9bb7b5d/web/handlers/admin/handler.go#L80-L82

And these two on public (`web/handlers`):

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/04824eeed78fedfd10b47e4e72081d18e9bb7b5d/web/handlers/http.go#L265-L266

cryptix:
Might need another pass after #105

PR #80, opened 2021-03-12
style the invite-list page

staltz:
I'm keeping this PR focused on just one page, I'll make other PRs for each other page. I'm not entirely confident about the layout I made here, when we find a potentially better layout, we can change it.

Screenshot from 2021-03-12 13-24-29


staltz:
PS @cryptix some ideas I stumbled upon while making this: (1) it would be useful to have a field "CreationDate" on the Invite model, so moderators on this page can know which invites are old (and probably could be revoked) and which are young, (2) I noticed that the alias suggestion doesn't have to be unique, maybe we could attach this invite logic to the alias database so that you cannot suggest an alias that is already taken (either by an actual alias registration, or taken in another invite as an alias suggestion). Ideas for others issues/PRs.

cryptix:
Yup, those are both good suggestions. The created time is a total no brainer and oversight.

Re alias suggestion: I thought of that but wanted to have the alias system in place first before doing this.

cryptix:
Re the layout: do we show the IDs on the other pages? I don't think users need to see those.

Other then that, it fits the vibe of the rest.

I think on the members list the add/create thing is under the heading?

These are the only inconsistencies I noticed.

staltz:
> Re alias suggestion: I thought of that but wanted to have the alias system in place first before doing this.

Oh right

> Re the layout: do we show the IDs on the other pages? I don't think users need to see those.

It's useful just to disambiguate the invite codes in case you created N invites with no aliases. If we have the CreationDate, we could remove the IDs.

> I think on the members list the add/create thing is under the heading?

I know, it's a different design here, it's a bit different concepts. Members list is about including known IDs, this one is about creation (where design references I've seen on Dribbble typically put outside of the table itself).

cblgh:
👍 

looks great :) does it work nicely on phones as well? i don't speak tailwind :~

cryptix:
oh and this updates #68, too :smile_cat:

cryptix (approved):

staltz:
> +1
> 
> looks great :) does it work nicely on phones as well? i don't speak tailwind :~

I'll check! Good reminder

staltz:
Ok, it's all broken on mobile screen sizes. I'll fix that

staltz:
Ok, now it's like this:

| Mobile | Desktop |
|----|----|
| Screen Shot 2021-03-12 at 16 55 48 | Screenshot from 2021-03-12 16-55-54 |

staltz:
Can haz merge? :cat2:

cryptix:
Very nice! Go ahead!

PR #88, opened 2021-03-16
Track when invites are created

### LINKED ISSUE #81: Track created date/time on invites
Raised in #80 

* [ ] Add a migration to add a column to the invites sqlite table
* [ ] Add a `CreatedAt time.Time` field to `roomdb.Invite`
* [ ] Changing the unpacking in `roomdb/sqlite/invites.go` to copy the field from the models to the roomdb type.

Some links:

* https://stackoverflow.com/questions/2614483/how-to-create-a-datetime-column-with-default-value-in-sqlite3
* https://sqlite.org/lang_datefunc.html

cryptix:
Fixes #81

created_at_relative

right now this uses [humanize.Time](https://pkg.go.dev/github.com/dustin/go-humanize?utm_source=godoc#Time) to turn the absolute time to a relative time from when the page is rendered.

As I'm writing this, I realize we need to either use absolutes or have a i18n-ified version of [this table](https://github.com/dustin/go-humanize/blob/9f541cc9db5d55bce703bd99987c9d5cb8eea45e/times.go#L47) and call `humanize.CustomRelTime` instead. but this magic can be hidden in the template func. Totally fine with doing that as a follow-up later once we can actually choose languages, tho.

* [x] Added bonus for having a tooltip with the absolute time.
To format the time inside the template you would do `{{.CreatedAt.Format "2006-01-02 15:04"}}` - where the constants come from the 1,2,3 (or 15 for 24hrs),4,5,6 enumerated style of [time.Fomat](https://pkg.go.dev/time#Time.Format). The _Example_ there has a much better explanation then the actual go doc string (don't ask me why).


staltz:
> Fixes #88

"Fixes itself" :smile:

staltz:
Awesome that you did the humanize plus absolute time tooltip! It's better than what I imagined (either humanize or absolute, not both).

cryptix:
i copy&:spaghetti:ed some tailwind thing into `web/style/input.css` to have a super simple tooltip.

with-tooltip_000

@staltz: if you feel like making the styling more consistent, i think this is good to merge.

cryptix:
You can take over this branch or we merge and you make a new one. I dont mind either way.

staltz:
Oh that's very nice! I'll just tweak some sizes and then we're all good. Thanks for doing the full-stack this time :)

staltz:
The `<table>` was getting cumbersome to work with, so I tried CSS Grid, and it was going well (but not a lot of visual advantages over `<table>`) until I realized that CSS Grid would destroy the `id`s we were using for tests, so I just dropped that whole approach, and tweaked some paddings instead. :P

staltz:
I also changed the date time format to ISO 8601 (internationally good). Wow, that Go `Time.Format` thing was really bonkers.

staltz (approved):

staltz:
@cryptix I'll let you press the green big button, for maximum satisfaction.

cryptix:
>  that Go Time.Format thing was really bonkers.

yea first time is not fun... thanks!

Issue #86, opened 2021-03-16
Add tests for JS client code to register and revoke aliases


cryptix:
Should be fairly simple to do after #85

PR #85, opened 2021-03-16
Use new ssb-room client module

cryptix:
Based on #78 so that should be merged first

Fixes #76

PR #95, opened 2021-03-23
add JS test for alias register and revoke

cryptix:
This is based on #85 which should be merged first.

I wanted to add a test for aliases to it but couldn't.

1) The client plugin needs a new version after https://github.com/ssb-ngi-pointer/ssb-room-client/pull/1
2) even with master of the above I get this:

```
/home/cryptix/ssb/ssb-room-client/lib/plugin-room-client.js:30
                rooms.get(roomKey).rpc.room.registerAlias(alias, sig, cb);

TypeError: Cannot read property 'registerAlias' of undefined
    at Object.registerAlias (/home/cryptix/ssb/ssb-room-client/lib/plugin-room-client.js:30:44)
    at Object.hooked [as registerAlias] (/home/cryptix/go-repos/ssb-rooms/muxrpc/test/nodejs/node_modules/hoox/index.js:10:15)
    at Timeout._onTimeout (/home/cryptix/go-repos/ssb-rooms/muxrpc/test/nodejs/testscripts/modern_aliases.js:12:23)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
```

I suspect some namespace/manifest meh-ness. I successfully used `room.registerAlias` already with `go-ssb/cmd/sbotcli` which ignores the manifest.


cryptix:
@staltz would be cool if you could take this.

I think this just needs an updated package.json in `muxrpc/test/nodejs`, updating the deps and running `go test -run Aliases` should then pass, I think.

staltz:
@cryptix I updated ssb-room-client (I think in the future I'd like to update secret-stack to support a nicer way of declaring remote APIs that are not implemented locally) and other dependencies, but now I think there are errors from Go:

```
revokeAlias: not your alias (moderators need to use the web dashboard of the room
```

Also, registerAlias should respond with a string matching the alias URL. This should also be asserted in tape.

muxrpc/test/nodejs/testscripts/modern_aliases.js
```diff
@@ -0,0 +1,25 @@
...
+  secretStackPlugins,
+  before,
+  after: (t, sbot, rpc, exit) => {
+
+    // give ssb-conn a moment to settle
+    setTimeout(() => {
+
+      sbot.roomClient.registerAlias(rpc.id, "alice", (err, ret) => {
+        t.error(err, 'registerAlias')
+        t.comment(`registerAlias value: ${ret}`)
```
staltz, line 14:
```suggestion
        t.ok(ret)
        t.equals(typeof ret, 'string')
        t.ok(new URL(ret))
```

cryptix:
Thank you! I can take this over again.

> I think there are errors from Go

There is a missing mock-setup call to have it not flunk out here:

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/07feacb5587e2baa912b8203909cf41fa6f0a672/muxrpc/handlers/alias/handler.go#L118

(background: these tests run with a mocked alias store so it doesn't actually store the alias)

cryptix:
All done and register now returns the URL.

Fixes #86

staltz (approved):
Looks great

PR #97, opened 2021-03-23
Update insert user

cblgh:
I updated the cli tool `insert-user` for a few reasons:
* there was a bug where you couldn't set moderators
* included more feedback when missing required arguments
* improved the error handling (it was hardwired to fail unless we input exactly 3 arguments, an assumption that was no longer valid :)

For convenience sake, I encoded a cool trick @cryptix showed me for generating a valid (but fake) ssb, in the bash script `generate-fake-id.sh`.

I also updated the readme to conform to the new behaviour.

### DIFF
--- a/cmd/insert-user/generate-fake-id.sh
+++ b/cmd/insert-user/generate-fake-id.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+id=$(dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0)
+echo "@$id.ed25519"
--- a/cmd/insert-user/generate-fake-id.sh
+++ b/cmd/insert-user/generate-fake-id.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 id=$(dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0)
-echo "@$id.ed25519"
+echo "@${id}.ed25519"
--- a/cmd/insert-user/main.go
+++ b/cmd/insert-user/main.go
@@ -65,11 +65,10 @@ func main() {
 
 		return nil
 	})
-
 	flag.Parse()
 
-	// we require at least 5 arguments: <executable> + -name <val> + -key <val>
-	//                               1              2     3       4    5
+	/* we require at least 5 arguments: <executable> + -name <val> + -key <val> */
+	/*                                  1              2     3       4    5     */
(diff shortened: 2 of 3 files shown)


cmd/insert-user/generate-fake-id.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+id=$(dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 -w0)
+echo "@$id.ed25519"
```
cryptix, line 3:
```suggestion
echo "@${id}.ed25519"
```

Just some bash-foo :upside_down_face:

cmd/insert-user/main.go
```diff
@@ -53,21 +68,25 @@ func main() {
 
 	flag.Parse()
 
-	if len(os.Args) != 3 {
-		fmt.Fprintf(os.Stderr, "usage: %s <user-name> <@theirPublicKey.ed25519>\n", os.Args[0])
-		flag.Usage()
-		os.Exit(1)
-		return
+	// we require at least 5 arguments: <executable> + -name <val> + -key <val>
+	//                               1              2     3       4    5
```
cblgh, line 72: lol go fmt was not kind to this line :crying_cat_face:

cmd/insert-user/main.go
```diff
@@ -53,21 +68,25 @@ func main() {
...
+	if len(os.Args) < 5 {
```
cryptix, line 73: I think this is fine but the comment tells me something is foul here.

There is https://pkg.go.dev/flag#Args which can be used to get the "non flag args", after flag.Parse() of course.

Obviously if we use it it would be zero and we still need a way to check that `-name` and `-key` were passed and counting the args doesn't tell us which those were either soo.. again. Might be fine.

Just putting it out there for future uses of the flag package.

cryptix (approved):
:+1: This documents much better what is required.

cmd/insert-user/main.go
cryptix: Sometimes It's more forgiving if you use `/*<newline> ... */`

cmd/insert-user/generate-fake-id.sh
cblgh: thanks :) definitely a novice in the bash apartment still :3

PR #99, opened 2021-03-24
Use name, not os.Args[1] when creating user

cblgh:
Haha, looks like I missed a part in my rewrite! This patch fixes insert-user to actually use the `name` variable, and not `os.Args[1]`.

### DIFF
--- a/cmd/insert-user/main.go
+++ b/cmd/insert-user/main.go
@@ -101,10 +101,10 @@ func main() {
 	}
 
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


cryptix (approved):
Tired review isn't good review, as got painstainkingly clear here.

For future references, I think it's fine if you merge these directly.

cblgh:
@cryptix gotcha! and i agree, this is the kind of pr that i would merge immediately in my own projects :)

PR #101, opened 2021-03-26
Style some login elements a bit

staltz:
## style the server-initiated ssb-http-auth template 

server-initiated-siwssb

## style the logged-in indicator to include PubKey.Ref

logged-in-corner


cryptix (commented):
Thanks! I already cherry-picked them into my working branch.

PR #103, opened 2021-03-26
style all the sign-in pages

staltz:
- I changed some Go
- I changed the sign-in colors from pink to green to be consistent with the signed-in indicator on the top-right corner
- Changed some routes a bit

| Page | Desktop | Mobile |
|----|----|----|
| `/login` | login-desktop | login-mobile |
| with SSB | withssb-desktop | withssb-mobile |
| Fallback | fallback-desktop | fallback-mobile |


cryptix (approved):
Gorgeous!

I wonder why the CI didn't run but I can rehabilitate the tests if you want. Should be mostly be pointing the password form requests at the right page.

cryptix:
The only other thing I'd remove is the block where it tries to figure out if `input` is feed or an alias in `web/handlers/auth/withssb.go`. I would have suggested a change but somewhy it's not possible to comment on regions outside of what is patched...

cryptix:
pulled into #90 :smile_cat:

PR #105, opened 2021-03-26
Add support for invite consumption via JSON

### LINKED ISSUE #71: Invites follow-up: JSON API endpoints
Right now we only have a HTTP/HTML interface, which includes CSRF token validation in production which would be cumbersome to implement for Apps.

Not 100% if we want #11 first or just mark better session handling as a TODO. I'm leaning on the latter so that we can make progress on this.

cryptix:
* [x] conform invites to https://github.com/ssb-ngi-pointer/rooms2/pull/35
* [x] change _accept_ page to _facade_
* [x] support JSON  consumption
* [x] support `x-www-form-urlencoded`
  * [ ] re-style pages (:grimacing:)
* [x] update consumption tests
* [x] re-consider CSRF bypass (It's fine, the invite code/token already is a secret)
* [x] remove member nickname / untangle alias suggestion

updates #13 

fixes #71


web/handlers/http.go
```diff
@@ -328,11 +333,26 @@ func New(
...
+	urlTo := web.NewURLTo(m)
+	consumeURL := urlTo(router.CompleteInviteConsume)
+
 	// apply HTTP middleware
 	middlewares := []func(http.Handler) http.Handler{
 		logging.InjectHandler(logger),
 		members.ContextInjecter(dbs.Members, authWithPassword, authWithSSB),
 		CSRF,
+
+		// this is an ugly hack to disable CSRF for certain requests that are done by apps
```
cryptix, line 345: I'm unhappy with this one but I'm not 100% sure if we _really_ need CSRF on this route since the token/inviteCode is a secret somehow already.

cryptix: The alternative would be to set a header on the `/join` request (`X-CSRF-TOKEN`) and have Apps send it back but that would need another change of the spec.

(See also: https://github.com/gorilla/csrf#javascript-applications)

arj03: If its like an API where you might have special authentication anyway then it should be fine to disable CSRF. Django has the same kind of escape hatch: https://docs.djangoproject.com/en/3.1/ref/csrf/#django.views.decorators.csrf.csrf_exempt

staltz:
@cryptix What's the difference between `invite/consumed.tmpl` and `invite/facade.tmpl`? Both seem to display the SSB URI.

staltz:
Oh, maybe I'm understanding now: is `CompleteInviteConsume` the `submissionUrl` which is POST-only? (If so, does it need an HTML page since it will only be called programmatically?)

cryptix:
> Oh, maybe I'm understanding now: is CompleteInviteConsume the submissionUrl which is POST-only? 

yes, consumed is _after_ they supplied their ID.

> If so, does it need an HTML page since it will only be called programmatically?

I kept the HTML form in case people can't get the URIs to work with their clients and want to input the pub manually. Although I agree that it is _advanced_ use, I'd still show the form if the handler isn't registered, together with a message that the ssb: handler isn't there.

staltz:
> I kept the HTML form in case people can't get the URIs to work with their clients and want to input the pub manually. Although I agree that it is _advanced_ use, I'd still show the form if the handler isn't registered, together with a message that the ssb: handler isn't there.

Oh, right, that's decent. I think I can make a separate page (just to declutter) for the form to input the `id`.

cryptix:
@staltz do you want to do the re-styling on this or a new PR? otherwise I think this is ready for review and possibly merge.

staltz:
@cryptix Separate PR.

This one is :heavy_check_mark: to merge :)

PR #107, opened 2021-03-29
Style dashboard

staltz:
I also checked that both desktop and mobile look good.

### DIFF
--- a/web/handlers/admin/handler.go
+++ b/web/handlers/admin/handler.go
@@ -65,28 +65,16 @@ func Handler(
 	mux.HandleFunc("/dashboard", r.HTML("admin/dashboard.tmpl", func(rw http.ResponseWriter, req *http.Request) (interface{}, error) {
 		onlineRefs := roomState.List()
 		onlineCount := len(onlineRefs)
-		members, err := dbs.Members.List(req.Context())
-		memberCount := 0
-		if err == nil {
-			memberCount = len(members)
-		}
-		invites, err := dbs.Invites.List(req.Context())
-		inviteCount := 0
-		if err == nil {
-			inviteCount = len(invites)
-		}
-		banned, err := dbs.DeniedKeys.List(req.Context())
-		bannedCount := 0
-		if err == nil {
-			bannedCount = len(banned)
-		}
+		memberCount := dbs.Members.Count(req.Context())
+		inviteCount := dbs.Invites.Count(req.Context())
+		deniedCount := dbs.DeniedKeys.Count(req.Context())
 		return struct {
 			OnlineRefs  []string
 			OnlineCount int
-			MemberCount int
-			InviteCount int
-			BannedCount int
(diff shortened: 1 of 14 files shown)


web/handlers/admin/handler.go
```diff
@@ -63,11 +63,30 @@ func Handler(
...
 
 	mux.HandleFunc("/dashboard", r.HTML("admin/dashboard.tmpl", func(rw http.ResponseWriter, req *http.Request) (interface{}, error) {
-		lst := roomState.List()
+		onlineRefs := roomState.List()
+		onlineCount := len(onlineRefs)
+		members, err := dbs.Members.List(req.Context())
+		memberCount := 0
+		if err == nil {
+			memberCount = len(members)
+		}
```
staltz, line 72: All this stuff should be converted to `.Count()` function on the interfaces. I tried to do that but got a bit confused with the generated Go code and how to handle all that.

web/handlers/admin/handler_test.go
```diff
@@ -19,9 +19,12 @@ func TestDashoard(t *testing.T) {
 	html, resp := ts.Client.GetHTML(url.String())
 	a.Equal(http.StatusOK, resp.Code, "wrong HTTP status code")
 
+	a.Equal(1, html.Find("#online").Size())
```
cryptix, line 22: was about to suggest this should but `.Length()` but TIL:

> Size(), which is an alias for Length()


You could test the actual content if you add the `Count()` func and use `ts.Members.CountReturns(23, nil)` to have it display 23.

web/handlers/admin/handler.go
staltz: Is there any utility in returning the error? I'm thinking that this method could just return zero in those cases.

cryptix: > Is there any utility in returning the error?

It borders on philosophical but I prefer to see those and not render the page instead of assuming no-one is connected, etc.

One possible error if a transaction bug sneaked is `sqlite: database is locked`.

cryptix: My general stance is, if errors aren't handled they need to be logged but if no-one is reading the log output that's more disservice of the software.

cryptix (approved):
I can live with ignoring [my suggestion](https://github.com/ssb-ngi-pointer/go-ssb-room/pull/107#discussion_r603326555) to get an error from Count(). We can add the error if we suspect a problem.

staltz:
> I can live with ignoring [my suggestion](https://github.com/ssb-ngi-pointer/go-ssb-room/pull/107#discussion_r603326555) to get an error from Count(). We can add the error if we suspect a problem.

I pressed resolve before your suggestion was made, so I didn't ignore it. I can do the error handling and logging.

staltz:
And my reasoning there was that `Count` seemed similar to `HasID` (which returns bool, not (bool,error)).

cryptix (approved):
:tada: :+1:

staltz:
Gonna fix that merge conflict, BRB

PR #120, opened 2021-04-07
use custom type marshalling via database/sql interface

cryptix:
* remove cruft from sqlite package
* update sqlboiler.toml
* fix default value in migration
* add default tests

### DIFF
--- a/roomdb/types.go
+++ b/roomdb/types.go
@@ -68,6 +68,32 @@ const (
 	ModeRestricted
 )
 
+// Implements the SQL marshaling interfaces (Scanner for Scan & Valuer for Value) for PrivacyMode
+
+// Scan implements https://pkg.go.dev/database/sql#Scanner to read integers into a privacy mode
+func (pm *PrivacyMode) Scan(src interface{}) error {
+	dbValue, ok := src.(int64)
+	if !ok {
+		return fmt.Errorf("unexpected type: %T", src)
+	}
+
+	privacyMode := PrivacyMode(dbValue)
+
+	err := privacyMode.IsValid()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


roomdb/types.go
```diff
@@ -68,6 +68,32 @@ const (
 	ModeRestricted
 )
 
+// Implements the SQL marshaling interfaces (Scanner for Scan & Valuer for Value) for PrivacyMode
+
+// Scan implements https://pkg.go.dev/database/sql#Scanner to read integers into a privacy mode
+func (pm *PrivacyMode) Scan(src interface{}) error {
+	dbValue, ok := src.(int64)
```
cryptix, line 75: this needs to know what the database stores it as internally.

roomdb/types.go
```diff
@@ -68,6 +68,32 @@ const (
...
+	}
+
+	privacyMode := PrivacyMode(dbValue)
+
+	err := privacyMode.IsValid()
+	if err != nil {
+		return err
+	}
+
+	*pm = privacyMode
```
cryptix, line 87: You had `pm = &privacyMode` which pleases compiler semantics (left type == right type) but doesn't effect the change you want.

That way it just overwrite the pointer in the scope of `Scan()` with a pointer to the value that was just created inside that function. As soon as the function returns, they will be gone.

The _receiver_ `*pm` already points to a valid `PrivacyMode`, it needs to be dereferenced so that we can update the value inside and overwrite the number from it's zero value to the one we got from the DB.

HTH

cblgh:
thanks for finishing this up @cryptix, i think i understand a bit more what i was missing in my attempt.

Issue #96, opened 2021-03-23
Implementing privacy modes


cryptix:
# Documentation
https://ssb-ngi-pointer.github.io/rooms2/#stakeholders
https://ssb-ngi-pointer.github.io/rooms2/#privacy-modes
https://ssb-ngi-pointer.github.io/rooms2/#participation

background:
https://github.com/ssb-ngi-pointer/go-ssb-room/issues/12

## implementation implications

The room server needs to support the three modes (open, community, restricted).

Areas where privacy mode needs to be considered:
* invites
* connection establishment
* Aliases: restricted rooms don't support aliases, because restricted means to be completely shut off from the public
* more?

## remaining
the following tasks are what remains. they may be implemented in #100 or in a separate follow-up pr

* [x] clean up thinking-out-loud comments
* [x] figure out which tests are low-hanging fruit for the privacy modes logic in #100
  * [x] write those tests
* [x] final review of #100, catching any last mistakes/bugs
* [x] go through rebase hell
* [x] in pr #126: implement setting the privacy mode from the dashboard / http frontend views
* [x] in pr #122: add deny list logic to the invites parts (yes?)
* [x] in pr #137: migrate privacy modes ui to a separate settings view (related to #134)
* [x] in pr #139: add test for deny list application on connection establishment, in `muxrpc/test/go`
* [x] in  pr #179: disable ui if privacy mode does not allow for actions (e.g. invites)
* [ ] in the other pr: write more tests 
  * [ ] verifying restricted mode behaviour
  * [ ] verifying community mode behaviour
  * [ ] verifying open mode behaviour
  * [ ] tests for changing the privacy mode over POST requests?
 
## ideas

First round: start server with a command line flag for the target privacy mode, expose in the interface what the current privacy mode is

:ballot_box_with_check:  add `type RoomConfig interface { ... }` to package roomdb but do not implement it in sqlite in the first pass of things.

 * to interface.go ( https://github.com/ssb-ngi-pointer/go-ssb-room/blob/master/roomdb/interface.go )

```go
type RoomConfig interface {
    GetPrivacyMode(context.Context) (PrivacyMode, error)
}
```

where `type PrivacyMode` similar to roomdb.RoleMember enum etc

* then implement RoomConfig interface in cmd/server
* then you can pass it into all the web/handlers.New(...) init code
[...]

PR #122, opened 2021-04-07
Expand denylist

cblgh:
This PR adds logic for denying invite consumption & connection establishment for banned keys (also called denied keys).

It's part of #96.

### DIFF
--- a/roomsrv/server.go
+++ b/roomsrv/server.go
@@ -65,8 +65,9 @@ type Server struct {
 
 	StateManager *roomstate.Manager
 
-	Members roomdb.MembersService
-	Aliases roomdb.AliasesService
+	Members    roomdb.MembersService
+	DeniedKeys roomdb.DeniedKeysService
+	Aliases    roomdb.AliasesService
 
 	authWithSSB       roomdb.AuthWithSSBService
 	authWithSSBBridge *signinwithssb.SignalBridge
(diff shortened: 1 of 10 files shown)


roomsrv/server.go
```diff
@@ -90,6 +92,7 @@ func New(
 	s.authorizer = membersdb
 
 	s.Members = membersdb
```
cblgh, line 94: @cryptix what's the point of having both s.authorizer & s.Members?

cblgh:
i think this is ready for review @cryptix! unsure if you also want to test connection establishment being denied, but i can't see any prior tests for package `roomsrv`so i assumed no.

roomsrv/server.go
cryptix: residual cruft, really. it can be removed now.

web/handlers/invites.go
```diff
@@ -190,11 +190,17 @@ func (h inviteHandler) consume(rw http.ResponseWriter, req *http.Request) {
...
 		}
 		newMember = *parsedID
-
 	default:
 		http.Error(rw, fmt.Sprintf("unhandled Content-Type (%q)", ct), http.StatusBadRequest)
 		return
 	}
+
+	// before consuming the invite: check if the invitee is banned
+	if h.deniedKeys.HasFeed(req.Context(), newMember) {
```
cryptix, line 199: :+1:

cryptix:
>  but i can't see any prior tests for package roomsrvso i assumed no.

yea, this would be a first but it would be good to add them still..

might be easier to add them to `muxrpc/test/go` since it's easier to create a keypair, ban it and connect there.

cblgh:
alright, i'll add that test in another PR! 👍

Issue #118, opened 2021-04-06
SSE for updating room dashboard state


cryptix:
who joined and left, etc. could be fed the same way as the login redirect dance and #113 will do it's magic.

Definitely a version 2.1 thing, though.

PR #126, opened 2021-04-08
Start styling privacy modes ui

cblgh:
mock:
68780Iz

todos:
* [x] translations
* [x] hooking up logic
* [x] stealing @staltz's dropdowny component
* [x] look into doing the grid stuff with tailwind

### DIFF
--- a/web/templates/admin/dashboard.tmpl
+++ b/web/templates/admin/dashboard.tmpl
@@ -88,13 +88,43 @@
   <div class="max-w-lg">
     <h2 class="text-xl tracking-tight font-bold text-black mt-2 mb-2">{{i18n "PrivacyModesTitle"}}</h2>
     <p class="mb-4">
-      {{i18n "ExplanationPrivacyModes"}} 
+      {{ i18n "ExplanationPrivacyModes" }} 
       <a class="text-purple-600 font-bold" href="https://ssb-ngi-pointer.github.io/rooms2/#privacy-modes">{{ i18n "RoomsSpecification" }}</a>.
     </p>
-    <select class="mb-2">
-      <option selected="">Community</option>
-    </select>
-    <div class="grid grid-cols-3 gap-y-2 mb-2">
+  <details class="mb-8 self-start w-96" id="change-privacy">
+    <summary class="px-3 py-1 w-96 rounded shadow bg-white ring-1 ring-gray-300 hover:bg-gray-100 cursor-pointer">
+    {{ i18n .CurrentMode.String }}
+    </summary>
+
+    <div class="absolute w-96 z-10 bg-white mt-2 shadow-xl ring-1 ring-gray-200 rounded divide-y flex flex-col items-stretch overflow-hidden">
+      {{ range .PrivacyModes }}
+          {{ if ne . $.CurrentMode }}
+          <form
+            action="/dashboard/set-privacy"
+            method="POST"
+            >
+            {{$.csrfField}}
+            <input type="hidden" name="privacy_mode" value="{{.}}">
+            <input
+              type="submit"
+              value="{{ i18n .String }}"
+              class="pl-3 pr-3 py-2 w-full text-left bg-white text-gray-700 hover:text-gray-900 hover:bg-gray-50 cursor-pointer"
+              />
+          </form>
+          {{ else }}
+          <div class="pr-3 py-2 text-gray-600 flex flex-row items-center cursor-default">
+            <div class="w-10 flex flex-row items-center justify-center">
+              <svg class="w-4 h-4" viewBox="0 0 24 24">
+                <path fill="currentColor" d="M21,7L9,19L3.5,13.5L4.91,12.09L9,16.17L19.59,5.59L21,7Z" />
+              </svg>
+            </div>
+            <span>{{ i18n .String }}</span>
+          </div>
+          {{end}}
+      {{end}}
+    </div>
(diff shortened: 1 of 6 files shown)


cblgh:
aiight this is what it looks like atm: 

cc @cryptix @staltz 
2021-04-09-135051_1280x890_scrot
2021-04-09-135128_1262x827_scrot

web/templates/admin/dashboard.tmpl
```diff
@@ -85,5 +85,54 @@
     </div>
     {{end}}
   </div>
+  <div class="max-w-lg">
+    <h2 class="text-xl tracking-tight font-bold text-black mt-2 mb-2">{{i18n "PrivacyModesTitle"}}</h2>
+    <p class="mb-4">
+      {{ i18n "ExplanationPrivacyModes" }} 
+      <a class="text-purple-600 font-bold" href="https://ssb-ngi-pointer.github.io/rooms2/#privacy-modes">{{ i18n "RoomsSpecification" }}</a>.
```
staltz, line 92:
```suggestion
      <a class="text-pink-600 underline" href="https://ssb-ngi-pointer.github.io/rooms2/#privacy-modes">{{ i18n "RoomsSpecification" }}</a>.
```

Just changing the link style to match the style of other "external" links.

web/templates/admin/dashboard.tmpl
```diff
@@ -85,5 +85,54 @@
...
+    </p>
+  <h3 class="text-gray-600 text-sm font-bold mb-2">{{ i18n "SetPrivacyModeTitle" }}</h3>
```
staltz, line 94:
```suggestion
  <h3 class="text-gray-400 text-sm font-bold mb-2">{{ i18n "SetPrivacyModeTitle" }}</h3>
```

To match the same style of the other dropdown in the member details page.

web/templates/admin/dashboard.tmpl
```diff
@@ -85,5 +85,54 @@
...
+          <form
+            action="{{urlTo "admin:dashboard:set-privacy" }}"
+            method="POST"
+            >
+            {{$.csrfField}}
+            <input type="hidden" name="privacy_mode" value="{{.}}">
+            <input
+              type="submit"
+              value="{{ i18n .String }}"
+              class="pl-3 pr-3 py-2 w-full text-left bg-white text-gray-700 hover:text-gray-900 hover:bg-gray-50 cursor-pointer"
```
staltz, line 112:
```suggestion
              class="pl-10 pr-3 py-2 w-full text-left bg-white text-gray-700 hover:text-gray-900 hover:bg-gray-50 cursor-pointer"
```

To horizontally align all the 3 texts "Open", "Community", "Restricted" regardless if they are selected or not. See line 117

staltz (changes requested):
@cblgh Good job! I have a few suggested commits for visual tweaks, and a bigger suggestion. Other than that, this looks visually good and the code refactor of `admin/dashboard.go` is great.

The bigger suggestion is this: the dashboard page should be reserved for high-level overview of the current state of the room, and most of the time there will be many (or non-zero) peers online, and each of those peers is rendered in a list below the counter. There are plans for making that list dynamic #118, and in that case, this privacy mode UI would jump up and down as peers join and leave, making it difficult to click on the buttons.

So I'd recommend we move all of this new UI to a new page, `Settings`, with its own left-side menu item, and for now there would only be the privacy mode setting. Note that this page would have to be guarded for permission access. Members cannot change the privacy mode.

web/templates/admin/dashboard.tmpl
cblgh: i noticed this, and chose 600 because i thought it was the better choice ;)

but i will commit your suggestion

staltz: We can actually change it (maybe 500?) in a separate PR, as long as they are consistent everywhere. My concern here was just to keep it consistent, I'm not adamant on 400.

cblgh:
@staltz thanks for the review! :)

regarding moving the logic to a separate page due to planned work, i would suggest that can be done when that work is actually in the pipeline? from #118, it seems that might be coming quite a bit down the road (and end up being left on the cutting floor, due to overriding priorities):

> @cryptix: Definitely a version 2.1 thing, though.

if this would have been signaled earlier, i would have taken that into account from the start :~

staltz:
> regarding moving the logic to a separate page due to planned work, i would suggest that can be done when that work is actually in the pipeline?

There are 3 reasons why to put it on a separate page: (1) #118, (2) Dashboard is for read-only high-level information, not for configuring the room, (3) room-wide configurations should be available only to admins and mods. You're right that (1) is future work, but (2) and (3) are currently applicable. If you need help, we can merge and I can do this page-moving work if you want.

> if this would have been signaled earlier, i would have taken that into account from the start :~

We could have communicated better, like from my part the spec uses "Dashboard" to mean "all the pages behind a login", while "Dashboard" in this repo is just one page with some counters.

cblgh:
Alright, well let's merge this then & perform the larger change in a separate PR 👍. Whether you do that or I do it doesn't really matter as much.

cblgh:
oh ya btw i tottttallly missed the part ranging over the IDs https://github.com/ssb-ngi-pointer/go-ssb-room/blob/ff028699d2cb490bc14f05970bea46260ace00c7/web/templates/admin/dashboard.tmpl#L80-L86

PR #136, opened 2021-04-09
fix server-initiated success redirect JS

### LINKED ISSUE #131: Server-initiated ssb-http-auth JS and SSE hiccups
On mobile, server-initiated ssb-http-auth works, but it's a bit fragile (sometimes works, sometimes needs a manual page refresh). I think that's because on mobile there is basically just one app "open and visible" at any single time, so once the browser switches to manyverse, maybe some logic on the browser goes to sleep, maybe it's some JS, maybe it's some SSE, I'm not sure. So when the user returns to the browser app, nothing happened (the page was not automatically redirected to the logged-in view, even though the muxrpc part of ssb-http-auth succeeded normally).

But we can probably fix this by doing some light JS on the server's page, where we listen to "window refocused" or something like that, to update state and do the redirect.

staltz:
Fixes #131. Turned out there was no logical or network problem, there was just a visual glitch: we need to do the page redirect only when the browser is focused. (Android's Firefox in this case, but should be a safe assumption for other browsers as well, desktop or mobile).

### DIFF
--- a/web/assets/login-events.js
+++ b/web/assets/login-events.js
@@ -23,6 +23,7 @@ evtSource.addEventListener('failed', (e) => {
   console.error(e.data);
 });
 
+// prepare for the case that the success event happens while the browser is not on screen.
 let hasFocus = true;
 window.addEventListener('blur', () => {
   hasFocus = false;
--- a/web/assets/login-events.js
+++ b/web/assets/login-events.js
@@ -23,8 +23,20 @@ evtSource.addEventListener('failed', (e) => {
   console.error(e.data);
 });
 
+let hasFocus = true;
+window.addEventListener('blur', () => {
+  hasFocus = false;
+});
+
 evtSource.addEventListener('success', (e) => {
   waitingElem.classList.add('hidden');
   evtSource.close();
-  window.location = `/withssb/finalize?token=${e.data}`;
+  const redirectTo = `/withssb/finalize?token=${e.data}`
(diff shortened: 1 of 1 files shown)


web/assets/login-events.js
```diff
@@ -23,8 +23,20 @@ evtSource.addEventListener('failed', (e) => {
   console.error(e.data);
 });
 
+let hasFocus = true;
```
cryptix, line 26:
```suggestion
// prepare for the case that the success event happens while the browser is not on screen.
let hasFocus = true;
```

web/assets/login-events.js
```diff
@@ -23,8 +23,20 @@ evtSource.addEventListener('failed', (e) => {
...
+});
+
 evtSource.addEventListener('success', (e) => {
   waitingElem.classList.add('hidden');
   evtSource.close();
-  window.location = `/withssb/finalize?token=${e.data}`;
+  const redirectTo = `/withssb/finalize?token=${e.data}`
+  if (hasFocus) {
+    window.location.replace(redirectTo);
+  } else {
```
cryptix, line 37:
```suggestion
  } else {
  // wait for the browser to be back in focus and redirect then
```

cryptix (approved):
SGTM just added two comments so that people know why this is here somewhen in the future.

staltz:
Thanks for the review!

PR #115, opened 2021-04-01
Better error messages

cryptix:
fixes #66 
fixes #23 

Most importantly: I added `errhandler.go` and `flashes.go` to `web/errors`, these use `localizeError()` which translates a few hardcoded error messages to the users locale. I updated the english defaults accordingly.

Most of the handlers now have `flashes weberrors.FlashHelper` which has three methods:

* `AddMessage(rw http.ResponseWriter, req *http.Request, label string)`
* `AddError(rw http.ResponseWriter, req *http.Request, err error)`
* `GetAll(rw http.ResponseWriter, req *http.Request) ([]FlashMessage, error)`

To render these there is a new base template `flashes.tmpl` which can be used with `{{ tempalte "flashes" .}}`, See the overviews of admin/members, admin/aliases, etc.

Noteworthy:
*  `go.mindeco.de/http/tester@v1.11.0` now handles cookies automatically which made it possible to simplify quite a bit of the testing code that uses login and csrf.

TODO:

* [x] style the flash messages (right now they are just red and green text)
* [x] flashify login errors
* [x] flashify notice edits

PR #143, opened 2021-04-13
Add language picker 

### LINKED ISSUE #45: Make language picking possible
There should be a way for visitors to pick their preferred language.

I think it would be okay to store these in the session cookie. This would use the `github.com/gorilla/sessions` store we already have here:

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/f774817536669c4f2ce2fe356a172d047ef5e780/web/handlers/http.go#L114

Then [Get](https://pkg.go.dev/github.com/gorilla/sessions#CookieStore.Get) the session (where name might be `room-settings`) and retreiev/set a "langauge value" from `session.Values`. See [the package documentation](https://pkg.go.dev/github.com/gorilla/sessions#pkg-overview) for some gotchas and maybe our [auth package](https://github.com/cryptix/go/blob/eb625e0824c4572ba1c7e10251c61ba1e34731b6/http/auth/auth.go#L132) for a concrete example. 

* [x] add a `/change-language` endpoint which updates the session
* [x] load the session and get the language in the helper code of our `web/i18n` package

For reference & a good way to get started with cookies, cryptix recommends the pairing:
> Just the cookie store get/set/save in handlers/auth/withssb AuthenticateRequest/Logout

This is follow up 3) of PR #40.

cblgh:
2021-04-13-163635_1309x1033_scrot
Closes #45 

tasks
* [x] for each translated language, use the language name as defined in its original language
  * [x] hack: add key `LanguageName` defining the language name of a given translation?
* [x] get the list of all currently translated languages (for iterating over in template)
* [x] add a route somewhere for accepting the `/change-language` POST request
* [x] do the cookie stuff per details in #45
* [x] tweak `web/i18n/helper.go` to query for any language cookies & use the corresponding language translation if the cookie is set
* [x] only render language picker when we have `>= 2` language translations to pick from
* [x] add default language setting to admin settings page
* [x] fix `/set-language` only working on non-admin pages (bizarre csrf error), fixed in bbd77b4
* [x] sort language options
* [x] add bool to settings table; column name `use_subdomain_for_aliases` with a default of `1`
* [x] test
  * [x] admin: set default language correctly sets language option in database
  * [x] user: posting `/set-language` sets a language cookie

### DIFF
--- a/roomdb/sqlite/roomconfig.go
+++ b/roomdb/sqlite/roomconfig.go
@@ -68,7 +68,6 @@ func (c Config) SetPrivacyMode(ctx context.Context, pm roomdb.PrivacyMode) error
 	return nil // alles gut!!
 }
 
-// TODO: use proper language tag from "golang.org/x/text/language"?
 func (c Config) GetDefaultLanguage(ctx context.Context) (string, error) {
 	config, err := models.FindConfig(ctx, c.db, configRowID)
 	if err != nil {
--- a/roomdb/sqlite/roomconfig.go
+++ b/roomdb/sqlite/roomconfig.go
@@ -40,7 +40,6 @@ func (c Config) SetPrivacyMode(ctx context.Context, pm roomdb.PrivacyMode) error
 		return err
 	}
 
-	// cblgh: a walkthrough of this step (again, now that i have some actual context) would be real good :)
 	err = transact(c.db, func(tx *sql.Tx) error {
 		// get the settings row
 		config, err := models.FindConfig(ctx, tx, configRowID)
@@ -57,6 +56,49 @@ func (c Config) SetPrivacyMode(ctx context.Context, pm roomdb.PrivacyMode) error
 		}
 		if rowsAffected == 0 {
 			return fmt.Errorf("setting privacy mode should have update the settings row, instead 0 rows were updated")
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	return nil // alles gut!!
+}
+
+// TODO: use proper language tag from "golang.org/x/text/language"?
+func (c Config) GetDefaultLanguage(ctx context.Context) (string, error) {
+	config, err := models.FindConfig(ctx, c.db, configRowID)
+	if err != nil {
+		return "", err
+	}
+
+	return config.DefaultLanguage, nil
+}
+
+func (c Config) SetDefaultLanguage(ctx context.Context, langTag string) error {
+	if len(langTag) == 0 {
+		return fmt.Errorf("language tag cannot be empty")
+	}
+
+	err := transact(c.db, func(tx *sql.Tx) error {
+		// get the settings row
+		config, err := models.FindConfig(ctx, tx, configRowID)
+		if err != nil {
+			return err
+		}
+
+		// set the new language tag
+		config.DefaultLanguage = langTag
+		// issue update stmt
+		rowsAffected, err := config.Update(ctx, tx, boil.Infer())
+		if err != nil {
+			return err
+		}
+		if rowsAffected == 0 {
+			return fmt.Errorf("setting default language should have update the settings row, instead 0 rows were updated")
 		}
 
 		return nil
--- a/web/handlers/admin/set_language_test.go
+++ b/web/handlers/admin/set_language_test.go
@@ -2,17 +2,16 @@ package admin
 
 import (
 	"net/http"
-  "strings"
+	"strings"
 	"testing"
 
 	"github.com/ssb-ngi-pointer/go-ssb-room/web/router"
 	"github.com/stretchr/testify/assert"
 )
 
-
 /* can't test English atm due to web/i18n/i18ntesting/testing.go:justTheKeys, which generates translations that are just
 * translationLabel = "translationLabel"
-*/
+ */
 // func TestLanguageSetDefaultLanguageEnglish(t *testing.T) {
 // 	ts := newSession(t)
 // 	a := assert.New(t)
@@ -40,6 +39,6 @@ func TestLanguageSetDefaultLanguage(t *testing.T) {
 	a.Equal(http.StatusOK, resp.Code, "Wrong HTTP status code")
 
 	summaryElement := html.Find("#language-summary")
-  summaryText := strings.TrimSpace(summaryElement.Text())
-  a.Equal("Deutsch", summaryText, "summary language should display german translation of language name")
+	summaryText := strings.TrimSpace(summaryElement.Text())
+	a.Equal("Deutsch", summaryText, "summary language should display german translation of language name")
 }
--- a/web/handlers/admin/set_language_test.go
+++ b/web/handlers/admin/set_language_test.go
@@ -1,0 +1,45 @@
+package admin
+
+import (
+	"net/http"
+  "strings"
+	"testing"
+
+	"github.com/ssb-ngi-pointer/go-ssb-room/web/router"
+	"github.com/stretchr/testify/assert"
+)
+
+
+/* can't test English atm due to web/i18n/i18ntesting/testing.go:justTheKeys, which generates translations that are just
+* translationLabel = "translationLabel"
+*/
+// func TestLanguageSetDefaultLanguageEnglish(t *testing.T) {
+// 	ts := newSession(t)
+// 	a := assert.New(t)
+//
+// 	ts.ConfigDB.GetDefaultLanguageReturns("en", nil)
+//
+// 	u := ts.URLTo(router.AdminSettings)
+// 	html, resp := ts.Client.GetHTML(u)
+// 	a.Equal(http.StatusOK, resp.Code, "Wrong HTTP status code")
+//
+//   fmt.Println(html.Html())
+// 	summaryElement := html.Find("#language-summary")
+//   summaryText := strings.TrimSpace(summaryElement.Text())
+//   a.Equal("English", summaryText, "summary language should display english translation of language name")
+// }
+
+func TestLanguageSetDefaultLanguage(t *testing.T) {
+	ts := newSession(t)
+	a := assert.New(t)
+
+	ts.ConfigDB.GetDefaultLanguageReturns("de", nil)
+
+	u := ts.URLTo(router.AdminSettings)
+	html, resp := ts.Client.GetHTML(u)
+	a.Equal(http.StatusOK, resp.Code, "Wrong HTTP status code")
+
+	summaryElement := html.Find("#language-summary")
+  summaryText := strings.TrimSpace(summaryElement.Text())
+  a.Equal("Deutsch", summaryText, "summary language should display german translation of language name")
+}
--- a/web/handlers/admin/settings.go
+++ b/web/handlers/admin/settings.go
@@ -115,13 +115,13 @@ func (h settingsHandler) getMember(w http.ResponseWriter, req *http.Request) *ro
 
 func (h settingsHandler) verifyPostRequirements(w http.ResponseWriter, req *http.Request) bool {
 	if req.Method != "POST" {
-		// TODO: proper error type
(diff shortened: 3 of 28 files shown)


staltz:
@cblgh Can you put this PR in draft mode to indicate its readiness?

web/i18n/defaults/active.en.toml
```diff
@@ -1,5 +1,8 @@
 # default localiztion file for english
 
+# the name of this language, in its language. Used for language picking
+LanguageName = "English"
```
cblgh, line 4: this is new key `LanguageName` is used for populating the language picker. it would be a good idea to document it alongside other instructions for adding translations to the existing room

cblgh:
this is the test failure i am seeing on my machine:
```
635.162497ms unit=tunnel called=endpoints
/home/cblgh/code/go/src/go-ssb-room/muxrpc/test/nodejs/testscripts/modern_aliases.js:12
      sbot.roomClient.registerAlias(rpc.id, "alice", (err, ret) => {
                      ^

TypeError: Cannot read property 'registerAlias' of undefined
    at Timeout._onTimeout (/home/cblgh/code/go/src/go-ssb-room/muxrpc/test/nodejs/testscripts/modern_aliases.js:12:23)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
level=debug t=5.541728553s event="network listen loop exited"
level=debug t=5.542631243s event="tunserv closing" msg="connections closed"
level=info t=5.542693387s event="tunserv closing" msg="closers closed"
--- FAIL: TestGoServerJSClientAliases (5.54s)
    setup_test.go:131: go server: @7j2jowwuHPQd5YRes3smYRKNrd3+uKteXdILPzjBfYQ=.ed25519
    setup_test.go:161: starting client alice
    setup_test.go:217: JS alice:1 @yVFwS8u/+Uiu6I1eJbfBaQtx70EVmrwnhSCmYO6Iy3k=.ed25519
    setup_test.go:199: node client alice: exited with exit status 1 (after 1.592460342s)
    setup_test.go:280: 
        	Error Trace:	setup_test.go:280
        	            				aliases_test.go:45
        	Error:      	Received unexpected error:
        	            	node client alice exited with exit status 1
        	Test:       	TestGoServerJSClientAliases
    aliases_test.go:47: 
        	Error Trace:	aliases_test.go:47
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestGoServerJSClientAliases
        	Messages:   	register call count
    setup_test.go:133: bot close: <nil>
```

cblgh:
2021-04-16-090351_1419x1059_scrot

working in parallel on the default language admin settings, it's uncovered some additions i need to make to the template function `list_languages` :)

cryptix:
re the test failure, i just checked out your branch and couldn't reproduce. I suspect the npm deps changed.

Try running `cd muxrpc/test/nodejs && npm ci && go test -run TestGoServerJSClientAliases`

staltz:
The test failure means that `ssb-room-client` wasn't installed properly, so just double check that it's in node_modules, the version is the latest, and in secret-stack we're doing the `.use(require('ssb-room-client'))`

cblgh:
ship this german translation PRONTO 
2021-04-16-155020_1256x639_scrot

cblgh:
ok, all that's remaining then is
* [ ] fix /settings/set-language only working on non-admin pages (bizarre csrf error)

cryptix:
Totally orthogonal but before we need yet another migration, can you add another bool to the settings table for #160 ?

column name `use_subdomain_for_aliases` and default it to `1`.

I will add the other helper functons on the roomdb interfaces myself later. Just don't want to touch that migration file again.

cblgh:
@cryptix gotcha! will do this after the current batch of work i'm doing :)

cblgh:
5a15cab look alright @cryptix?

roomdb/sqlite/roomconfig.go
```diff
@@ -68,3 +67,46 @@ func (c Config) SetPrivacyMode(ctx context.Context, pm roomdb.PrivacyMode) error
 
 	return nil // alles gut!!
 }
+
+// TODO: use proper language tag from "golang.org/x/text/language"?
```
cryptix, line 71: Good suggestion but lets make this a follow up issue for when we actually receive translations.

web/handlers/admin/set_language_test.go
```diff
@@ -0,0 +1,45 @@
...
+	a := assert.New(t)
+
+	ts.ConfigDB.GetDefaultLanguageReturns("de", nil)
+
+	u := ts.URLTo(router.AdminSettings)
+	html, resp := ts.Client.GetHTML(u)
+	a.Equal(http.StatusOK, resp.Code, "Wrong HTTP status code")
+
+	summaryElement := html.Find("#language-summary")
+  summaryText := strings.TrimSpace(summaryElement.Text())
```
cryptix, line 43: gofmt please

web/handlers/admin/settings.go
```diff
@@ -78,6 +90,44 @@ func (h settingsHandler) setPrivacy(w http.ResponseWriter, req *http.Request) {
...
+		h.r.Error(w, req, http.StatusInternalServerError, err)
+		return nil
+	}
+	return currentMember
+}
+
+func (h settingsHandler) verifyPostRequirements(w http.ResponseWriter, req *http.Request) bool {
+	if req.Method != "POST" {
+		// TODO: proper error type
+		h.r.Error(w, req, http.StatusBadRequest, fmt.Errorf("bad request"))
```
cryptix, line 119:
```suggestion
		err := weberrors.ErrBadRequest{Where: "HTTP Method", Details: fmt.Errorf("expected POST not %s", req.Method)}
		h.r.Error(w, req, http.StatusBadRequest, err)
```

these landed in #115. :)

ps: i like this helper, could possible detatch it from `settingsHelper` and use it in the other handlers, too but let's do that in a later refactor.

web/handlers/admin/settings.go
```diff
@@ -78,6 +90,44 @@ func (h settingsHandler) setPrivacy(w http.ResponseWriter, req *http.Request) {
...
+		return false
+	}
+	if err := req.ParseForm(); err != nil {
+		// TODO: proper error type
+		h.r.Error(w, req, http.StatusBadRequest, fmt.Errorf("bad request: %w", err))
```
cryptix, line 124:
```suggestion
		err = weberrors.ErrBadRequest{Where: "Form data", Details: err}
		h.r.Error(w, req, http.StatusBadRequest, err)
```

web/handlers/http.go
```diff
@@ -122,6 +123,42 @@ func New(
...
+			return func() int {
+				return len(locHelper.ListLanguages())
+			}
+		}),
+
+		render.InjectTemplateFunc("list_languages", func(r *http.Request) interface{} {
+			csrfElement := csrf.TemplateField(r)
+
+			createFormElement := func(postRoute, tag, translation, classList string) string {
+				return fmt.Sprintf(`
```
cryptix, line 136: I'd invite you to try this as a standalone template. 6 format inputs is quite a lot, i think.

I'd place this above the `render.Options` (or somehwere sensible, this func is getting quite long..)


```go
type languageChangeTemplateData struct {
	Action    string
	CSRFField template.HTML
	Language  string
	Page      string
	FormValue string
	ClassList string
}

languageChangeForm, err := template.New("name").Parse(`
            <form
              action="{{.Action}}"
              method="POST"
              >
              {{.CSRFField}}
              <input type="hidden" name="lang" value="{{.Language}}">
              <input type="hidden" name="page" value="{{.Page}}">
              <input
                type="submit"
                value="{{.FormValue}}"
                class="{{.ClassList}}"
                />
            </form>
`)
if err!=nil { ... }
```

And then just call this inside the `list_languages` template func:

```go
	var (
		buf       bytes.Buffer
		languages = locHelper.ListLanguages()
	)

	for _, entry := range languages {

		data := languageChangeTemplateData{
			Action:    postRoute,
			CSRFField: csrfElement,
			Language:  tag,
			Page:      r.RequestURI,
			FormValue: translation,
			ClassList: classList,
		}
		err = languageChangeForm.Execute(&buf, data)
		if err != nil {
			return "", err
		}
	}

	return template.HTML(buf.String()), nil

```

ps: always writing into `buf` will also save us from a medium expensive/wastefull "small strings onto one big one".

web/handlers/http.go
```diff
@@ -237,6 +276,26 @@ func New(
...
 	mainMux.Handle("/admin/", members.AuthenticateFromContext(r)(adminHandler))
 
+	// handle setting language
+	m.Get(router.CompleteSetLanguage).HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		lang := req.FormValue("lang")
+		previousRoute := req.FormValue("page")
+
+		session, err := cookieStore.Get(req, i18n.LanguageCookieName)
+		if err != nil {
+			fmt.Errorf("cookie error? %w\n", err)
```
cryptix, line 286:
```suggestion
			errHandler.Error(w, req, http.StatusInternalServerError,err)
```

web/handlers/http.go
```diff
@@ -237,6 +276,26 @@ func New(
...
+			return
+		}
+
+		session.Values["lang"] = lang
+		err = session.Save(req, w)
+		if err != nil {
+			fmt.Errorf("we failed to save the language session cookie %w\n", err)
```
cryptix, line 293:
```suggestion
			err = fmt.Errorf("we failed to save the language session cookie %w\n", err)
			errHandler.Error(w, req, http.StatusInternalServerError, err)
			return
```

web/handlers/invites_test.go
```diff
@@ -131,7 +131,7 @@ func TestInviteConsumeInviteHTTP(t *testing.T) {
 	})
 
 	// get the corresponding token from the page
-	csrfTokenElem := doc.Find("input[name='gorilla.csrf.Token']")
+	csrfTokenElem := doc.Find(`form[action="/invite/consume"] input[name="gorilla.csrf.Token"]`)
```
cryptix, line 134: maybe stick an `#id` on the form, these links might break some day if we move the routes around.

web/handlers/notices_test.go
```diff
@@ -92,7 +92,7 @@ func TestNoticesEditButtonVisible(t *testing.T) {
 	csrfCookie := resp.Result().Cookies()
 	a.True(len(csrfCookie) > 0, "should have one cookie for CSRF protection validation")
 
-	csrfTokenElem := doc.Find("input[type=hidden]")
+	csrfTokenElem := doc.Find(`form[action="/fallback/finalize"] input[type="hidden"]`)
```
cryptix, line 95: see above.

web/handlers/set_language_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	// "github.com/ssb-ngi-pointer/go-ssb-room/roomdb"
```
cryptix, line 12:
```suggestion
```

cryptix (changes requested):
Some nagging but really like the substance! Good for merge after these changes.

web/handlers/http.go
cryptix: the `data := languageChangeTemplateData{ ... }` surely needs to be changed a little but I think you get the idea.

cblgh:
Thanks for the code review @cryptix! I'll get on implementing the remaining changes & fixing the tests :)

web/handlers/http.go
cblgh: wdyt about the implementation in 546b249? :)

cryptix (approved):
Very nice now! LGTM

PR #144, opened 2021-04-13
change withssb login QR code to be a login URL

### LINKED ISSUE #132: ssb-http-auth QR Code should contain URL
I tested the ssb-http-auth server-initiated QR code in 3 popular QR code scanner apps on Android and they didn't work (well, they correctly detected the contents of the QR to be the SSB URI, but they didn't attempt to open the SSB URI using a compatible app), so it seems they'll be useless like that.

We could change the contents of the QR code to be *current URL*, which means that once scanned, the scanner app would redirect to the browser *on the phone*, which is different than what we we're aiming at with these QR codes (we wanted to login on desktop using mobile SSB ID), but at least it will successfully login on mobile, instead of just showing you the SSB URI as raw text, which is not really useful to an end-user if they can't "activate" that.

staltz:
Fixes #132


web/handlers/auth/withssb.go
```diff
@@ -357,26 +366,42 @@ func (h WithSSBHandler) serverInitiated() (templateData, error) {
...
 
-	// generate a QR code with the token inside so that you can open it easily in a supporting mobile app
-	qrCode, err := qrcode.New(startAuthURI.String(), qrcode.Medium)
-	if err != nil {
-		return templateData{}, err
-	}
+	var qrURI string
+	if isSolvingRemotely {
+		qrURI = ""
+	} else {
```
cryptix, line 372:
```suggestion
	if !isSolvingRemotely {
```

`qrURI` doesnt need to be initiallized manually if we don't use it (empty string is the zero value for it anyhow).

cryptix (approved):
SGTM, one style nitpick.

staltz:
Nice, let's merge

PR #146, opened 2021-04-14
fix invite QR code

staltz:
The claim-invite QR code should not be the SSB URI (because most mobile scanners don't support it), it should be a web URL.


cryptix (approved):
as discussed, SGTM

PR #155, opened 2021-04-15
more content and member links on the dashboard

### LINKED ISSUE #153: On dashboard, list of members should show aliases
- [ ] If the member has aliases, then instead of their SSB ID, render their *1st alias*
- [ ] The member should be clickable, going to "member details" page

staltz:
Closes #153 and #147 

Screenshot from 2021-04-15 16-10-44

### DIFF
--- a/web/handlers/admin/dashboard.go
+++ b/web/handlers/admin/dashboard.go
@@ -7,7 +7,10 @@ import (
 	"net/http"
 
 	"go.mindeco.de/http/render"
+	refs "go.mindeco.de/ssb-refs"
 
+	"github.com/ssb-ngi-pointer/go-ssb-room/internal/network"
+	"github.com/ssb-ngi-pointer/go-ssb-room/roomdb"
 	"github.com/ssb-ngi-pointer/go-ssb-room/roomstate"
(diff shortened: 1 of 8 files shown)


web/handlers/admin/dashboard.go
```diff
@@ -17,31 +20,52 @@ type dashboardHandler struct {
...
 
 func (h dashboardHandler) overview(w http.ResponseWriter, req *http.Request) (interface{}, error) {
+	roomRef := h.netInfo.RoomID.Ref()
+
 	onlineRefs := h.roomState.List()
-	onlineCount := len(onlineRefs)
+
+	onlineMembers := make([]roomdb.Member, len(onlineRefs))
+	for i := range onlineRefs {
+		ref, err := refs.ParseFeedRef(onlineRefs[i])
```
cryptix, line 34: fine for now. Should probably change the `roomState.List()` to return `FeedRef`s but can happen later.

cryptix (approved):

web/handlers/admin/dashboard.go
staltz: Yeah I thought about it. I can add a TODO

PR #159, opened 2021-04-16
show admin and moderator tags on members list

### LINKED ISSUE #148: On members list, render who is admin / mod

staltz:
Closes #148 

Screenshot from 2021-04-16 13-08-13

### DIFF
--- a/web/handlers/admin/members_test.go
+++ b/web/handlers/admin/members_test.go
@@ -107,9 +107,9 @@ func TestMembers(t *testing.T) {
 	}
 	ts.MembersDB.ListReturns(lst, nil)
 
-	membersOveriwURL := ts.URLTo(router.AdminMembersOverview)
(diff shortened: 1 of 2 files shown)


cryptix:
Using the numbers seems a bit bad.

Alternatively you could do `{{if eq .Role.String RoleMember}}`.

Not sure if it needs a test but would be bad to regress on.

cryptix:
Using the numbers seems a bit bad.

Alternatively you could do `{{if eq .Role.String RoleMember}}`.

Not sure if it needs a test but would be bad to regress on.

staltz:
Great! I actually tried `{{if eq .Role "RoleAdmin"}}` and it didn't work. The trick was `.Role.String` :ok_hand: 

(Updated the PR)

staltz:
Added tests

cryptix (approved):
nice! thanks for adding the tests, too.

SGTM

Issue #152, opened 2021-04-15
alias URL should be stable when visited


staltz:
Visiting `alice.room.com` should not redirect to (i.e. change the browser's address bar) `room.com/aliases/alice`, it should stay on `alice.room.com` to allow copy-pasting it.

cryptix:
cc #117

cryptix:
I agree but shouldn't we expand this to generate the right URL scheme?

I'd really like to support the Tor hidden-service scheme where wildcard domains are not possible.

staltz:
I agree with that, but isn't it a separate issue? (Or separate in the sense that it could be done later, and that there are different strategies how to support Tor hidden-service)

cryptix:
True, just related because you mentioned copy and paste and for me the first step is copying it from somewhere.


I just had a shower thought to fix the issue on nginx. Instead of URL rewrite, which by all accounts is not possible when changing the domain, we could have a 2nd proxy_pass just for the aliases.

Issue #117, opened 2021-04-05
Hosting documentation: HTTP(S) Proxy


cryptix:
We currently assume a standard HTTPS server in front of go-ssb-room to facilitate TLS termination and certificate management. This should be possible with most modern HTTP servers since it's a pretty standard practice, known as [reverse proxying](https://en.wikipedia.org/wiki/Reverse_proxy).

Two bits of rational:

1) People usually want to have more then one site on their server. Put differently we could have LetsEncrypt inside the go-ssb-room server but it would have to listen on port :443, blocking the use of other domains on the same IP.
2) Listening on :443 can be pretty annoying (you might need root priviliges or similar capabilities).

go-ssb-room needs three headers to function properly, which need to be forwarded by the webserver.

* `X-Forwarded-Host` as which domain name the room is running (enforce strict TLS checking)
* `X-Forwarded-Proto` to ensure that TLS is used (and redirect if necessary)
* `X-Forwarded-For` the remote IP address of the client accessing the room (used for rate limiting)

Here are three examples using common sever software:

# Nginx

**TODO: Needs websocket and error handlers**

https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/

Here is what we are using on hermies.club:

```
server {
    server_name hermies.club;

    listen 443 ssl; # managed by Certbot

    ssl_certificate /etc/letsencrypt/live/hermies.club/fullchain.pem; # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/hermies.club/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot

    location / {
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-For $remote_addr;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_pass http://localhost:8899;
    }

    # TODO: https://nginx.org/en/docs/http/websocket.html
    # TODO: https://blog.tarq.io/nginx-catch-all-error-pages/
}

server {
    server_name "~^(?<alias>\w+)\.hermies\.club$";

    listen 443 ssl; # managed by Certbot

    ssl_certificate /etc/letsencrypt/live/hermies.club/fullchain.pem; # managed by Certbot
[...]

PR #162, opened 2021-04-16
allow wildcard of the configured domain, too

cryptix:
Fix for #152 when used with the new config in #117.


staltz (approved):
I took a look at all files. Looks solid! I made issue #168 as a follow-up, just a reminder, really.

PR #163, opened 2021-04-16
refactor templates that have ssb uri links

staltz:
This is just a refactor PR in preparation of styling alias pages.


web/templates/invite/facade.tmpl
```diff
@@ -6,13 +6,12 @@
 
   <a
     id="join-room-uri"
-    href="#"
-    data-href="{{.JoinRoomURI}}"
+    href="{{.JoinRoomURI}}"
```
staltz, line 9: Using `href` instead of `data-href` allows users to right-click and "Copy link", and it seems that the JS behavior remains the same after this change (tested on desktop and mobile browsers).

staltz:
Fixed tests

cryptix (approved):
SGTM

PR #191, opened 2021-04-22
improve sign-in with ssb JS for Firefox

staltz:


cryptix (approved):
So great we came up with this trick!

PR #193, opened 2021-04-22
dont allow upercase chars in aliases (fix #192)

### LINKED ISSUE #192: alias registration should refuse uppercase characters

cryptix:
Thanks for testing @soapdog!

Using strings with UpperCase doesn't work in Firefox because turns it into `uppercase.$domain` which then is not found because it matches case sensitive...

### DIFF
--- a/internal/aliases/names.go
+++ b/internal/aliases/names.go
@@ -2,16 +2,12 @@
 
 package aliases
 
-import (
-	"fmt"
-)
-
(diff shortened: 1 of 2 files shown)


staltz:
> doesn't work in Firefox

Yeah and also Chrome, I tested

staltz (approved):
## :taco:

Issue #190, opened 2021-04-22
muxrpc client crashes when go-ssb-room closes or something


staltz:
## Reproduction steps

- Alice uses Manyverse to connect to hermies.club
- Alice is connect to other peers there via tunnel.connect
- hermies.club got hard-restarted to update its code
- Alice's Manyverse crashes with the error below:

```
{ name: 'Error', message: 'muxrpc: session terminated', stack: '' }
```

Not a lot more info is given, no stacktrace or anything. It might be a bug in JS ssb-conn (or muxrpc or elsewhere), but at least it would be good to know exactly what causes this.

cryptix:
Some thoughts: https://github.com/ssb-ngi-pointer/go-ssb-room/issues/170#issuecomment-823423171

staltz:
I'm keeping my manyverse open and connected to hermies.club and this crash occurs frequently. I got another log this time, and it has more useful information:

```
04-22 20:08:22.240 26309 26555 E NODEJS-MOBILE: 2021-04-22T17:08:22.240Z ssb:room-client we wish to connect to { name: 'tunnel', portal: '@uMYDVPuEKftL4SzpRGVyQxLdyPkOiX7njit7+qT/7IQ=.ed25519', target: '@p13zSAiOpguI9nsawkGijsnMfWmFd5rlUNpzekEE+vI=.ed25519' }
04-22 20:08:22.240 26309 26555 E NODEJS-MOBILE: 2021-04-22T17:08:22.240Z ssb:room-client will call tunnel.connect at @p13zSAiOpguI9nsawkGijsnMfWmFd5rlUNpzekEE+vI=.ed25519 via room @uMYDVPuEKftL4SzpRGVyQxLdyPkOiX7njit7+qT/7IQ=.ed25519
[...]

PR #201, opened 2021-04-26
Add a test where clients are disconnected forcefully

cryptix:
This new test is just like the normal _make a tunnel through the server with two js clients_-test.

The only difference is: just after the tunnel is build and the clients are about to sign off happily, the server cuts the connection (by simply shutting down).

With the bug we isolated in #190, this leads the nodejs processes to exit immediately because of an uncaught throw in pull-stream drain.

To make these tests pass ssb-room-client needs to be patched [here](https://github.com/ssb-ngi-pointer/ssb-room-client/blob/5bab562e080582d8892fb1b1807f7c3e0af4f072/src/room-observer.ts#L46) to have an extra done callback on the drain call.

Additionally we advise for pull-stream to have a better way of signaling the inability to handle the error then just [throwing the error from the network](https://github.com/pull-stream/pull-stream/blob/603b7dc5c53569cb4abe2e8953841b5d370f3f71/sinks/drain.js#L22) without an indication who called drain without the optional 2nd callback. @staltz gave me this idea and I managed to find the culprit in room-observer.ts with it:

```patch
diff --git a/sinks/drain.js b/sinks/drain.js
index 2025462..ef8c6fc 100644
--- a/sinks/drain.js
+++ b/sinks/drain.js
@@ -3,6 +3,11 @@
 module.exports = function drain (op, done) {
   var read, abort
 
+  var causeErr
+  if (typeof done === 'undefined') {
+    causeErr = new Error('no done callback supplied')
+  }
+
   function sink (_read) {
     read = _read
     if(abort) return sink.abort()
@@ -18,8 +23,10 @@ module.exports = function drain (op, done) {
             if(end = end || abort) {
               loop = false
               if(done) done(end === true ? null : end)
-              else if(end && end !== true)
+              else if(end && end !== true) {
+                console.warn(causeErr)
                 throw end
+              }
             }
             else if(op && false === op(data) || abort) {
               loop = false
```

TODO: publish updated nodejs deps and update our package.json to use them.


staltz:
I'm working on a fix to ssb-room-client, will let you know when you can update the version here

staltz:
Ship it

staltz (approved):

staltz:
# :heavy_check_mark:

PR #207, opened 2021-05-10
Bugfixing

### LINKED ISSUE #176: Members can create notices
I am logged in with a user that has RoleMember, and I try to create a notice. Instead of running into an error, I am able to create the notice.

Even if we disable the UI, I feel this should still be handled by a check in the server as well.

cryptix:
* ( 2e5af65 ) assure tunnel.connect doesn't work on self 
* ( 3f6bc2b ) update muxrpc to v2.0.4 fixes for manifest and aborting streams
* ( 990ea7e & e9ee232) fix #176 
* ( cf3cf17 ) remove `tunnel.isRoom` and `tunnel.endpoints` call logging


staltz (approved):
Looks good!

PR #209, opened 2021-05-10
have a clear error on taken aliases

### LINKED ISSUE #194: Alias registration muxrpc errors could be more clear
Instead of this:

```
{
  name: 'Error',
  message: 'registerAlias: could not register alias: transaction failed, rolling back: models: unable to insert into aliases: UNIQUE constraint failed: aliases.name',
  stack: ''
}
```

the client could receive something like `this alias is already taken`. Something that could be easier for apps to parse and then render an appropriate and localized popup on the UI.

Note, not high priority.

cryptix:
fixes #194 - It will now show as:

> transaction failed, rolling back: roomdb: alias (\"thealias\") is already taken`

### DIFF
--- a/muxrpc/handlers/alias/handler.go
+++ b/muxrpc/handlers/alias/handler.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


staltz:
To a muxrpc JS client, will it look like 

`transaction failed, rolling back: roomdb: alias ("thealias") is already taken`

or like

`alias ("thealias") is already taken`

?

cryptix:
It would show as:

> registerAlias: could not register alias: transaction failed, rolling back: roomdb: alias ("thealias") is already taken`

I could slim this case to the following tho, if you prefer. 

>  roomdb: alias ("thealias") is already taken

staltz:
Yeah it would be best if it's the latter, it's shorter, and the `transaction failed, rolling back` seems like the server is leaking irrelevant information to the client. I would even drop the `roomdb: ` if possible, for similar reasons. Otherwise, the code looks good and :+1: to have tests for this.

staltz (approved):
Thanks @cryptix!

Issue #116, opened 2021-04-02
Support checking passwords against haveibeenpwnd


cryptix:
during our demo yesterday @arj03 raised the concern that fallback passwords might be brute-forced by some means.

While we do have rate-limiting already, stuffing of leaked passwords is a valid concern still. Therefore, once users can update and reset their own passwords we probably want to add support for the Pwned passwords lookup: https://haveibeenpwned.com/API/v2#PwnedPasswords

tl;dr: you hash the new password of your user, send them a prefix of the hash and the result is all the matches of known passwords and the room can then check if the rest of the hash is in the list of returned values.

cryptix:
Probably will use https://github.com/mattevans/pwned-passwords for this.

decentral1se:
afaui this is done already? super nice feature.

Issue #98, opened 2021-03-23
Allow people to (re)set their fallback password


cryptix:
Right now the only way to create a fallback password login is with `cmd/insert-user` (ie. shell access to the system hosting the room).

We definitely want a way that doesn't require that. @cblgh suggested a button for admins to generate a reset link which then can be passed to the user so that they can reset it themselves and I agree that it's a valid solution.

Issue #93, opened 2021-03-22
follow-up: change ones own fallback password


cryptix:

decentral1se:
looks done also via `/admin/member?id=<int>` :partying_face: 

![image](https://user-images.githubusercontent.com/1991377/162607759-ea3044dc-ef34-4ef0-87ff-c144bf629638.png)

PR #211, opened 2021-05-11
Overhaul passwords

cryptix:
# TODO

* [x] make it possible to change ones own password (#93)
* [x] add a way to reset passwords (#98)
* [x] check new passwords against HIBP (#116)
* [x] replace _fallback login_ with `@ssbid || an alias`
* [x] ~~Update privacy notice to indicate use of HIBP.~~ See #214 
* [x] Add tests
  * [x] Use form to update own password
  * [x] Create Reset Link
  * [x] Consume link to update password (not logged in)

## add label to own entry in members list ( eaaed06 )

Shows a green "this is you" label on the members list. Green because that's also the color of the tag on the top right. Also changes the color of the  moderator label from green to blue to be less confusing. (ps: totally open for another color then blue for mods, just didnt want both to be green)

this-is-you

cc @SoapDog IIRC you suggested this on our office hour. I really liked the idea.


web/templates/base.tmpl
```diff
@@ -35,7 +35,10 @@
           <svg class="text-green-500 w-4 h-4 mr-1" viewBox="0 0 24 24">
             <path fill="currentColor" d="M22,18V22H18V19H15V16H12L9.74,13.74C9.19,13.91 8.61,14 8,14A6,6 0 0,1 2,8A6,6 0 0,1 8,2A6,6 0 0,1 14,8C14,8.61 13.91,9.19 13.74,9.74L22,18M7,5A2,2 0 0,0 5,7A2,2 0 0,0 7,9A2,2 0 0,0 9,7A2,2 0 0,0 7,5Z" />
           </svg>
-          <span class="text-green-500 text-sm truncate w-32">{{$user.PubKey.Ref}}</span>
+          <!-- TODO: i'd like to include the key SVG in the lnk but this breaks the layout -->
```
cryptix, line 38: cc @staltz

staltz:
> Shows a green "this is you" label on the members list. Green because that's also the color of the tag on the top right. Also changes the color of the moderator label from green to blue to be less confusing. (ps: totally open for another color then blue for mods, just didnt want both to be green)

Yep, this is a great idea!

> make it possible to change ones own password

I just had an idea: what if we allow the "username" field to be either the SSB ID or one of the aliases?

cryptix:
> what if we allow the "username" field to be either the SSB ID or one of the aliases?

Aaah..! I see, yea that works, too! I like that, will add it to the list.

cryptix:
Lot's of new stuff in here. Did all the things but need to add tests for the new flow.

cryptix:
I was tempted to also add password complexity checks other then length to the server side but would prefer if we add something like [zxcvbn](https://github.com/dropbox/zxcvbn) to the fronted directly.

cblgh:
hahaha holy shit that project name is so bad & so good

cmd/insert-user/main.go
```diff
@@ -65,18 +51,16 @@ func main() {
...
 	flag.Parse()
 
-	/* we require at least 5 arguments: <executable> + -login <val> + -key <val> */
-	/*                                  1              2     3       4    5     */
-	if len(os.Args) < 5 {
-		cliMissingArguments("please provide the default arguments -login and -key")
+	/* we require at least 3 arguments: <executable>  <pub_key> */
+	/*                                  1              2        */
+	if len(flag.Args()) != 1 {
+		cliMissingArguments("please provide the a public key")
```
cblgh, line 57:
```suggestion
		cliMissingArguments("please provide a public key")
```

cryptix:
> that project name 

I had a hunch it's up your ally... ;)

staltz (approved):
Looks good. I didn't test it out in the browser but from the templates I have an imagination of what it looks like. Good that this PR was full stack and had tests too.

Issue #199, opened 2021-04-26
Test open privacy mode behaviour


cblgh:
Related to & waiting for #102

cryptix:
I did a lot of the open test in #215 but need to double check if we need to make muxrpc related changes.

cryptix:
re my previous comments: I think this might be done.

- membership on new connections are checked in restricted
- `tunnel.endpoints` only works for members in community and restricted.

PR #216, opened 2021-05-14
Add open invites

### LINKED ISSUE #102: follow-up to privacy modes: open invites
initially part of #96 but while talking with @cblgh about #100 I realized that the invites implementation needs to be overhauled to conform with https://github.com/ssb-ngi-pointer/rooms2/pull/35 first anyhow.

I _suspect_ we don't strictly need _invite creation_ (which is currently tied to members here) but basically have open consumption that makes people internal users no matter what the `inviteCode`.

So in open mode we could have a `Create Invite` button _somewhere_ that just returns a link with a random code.

cryptix:
fixes #102 
updates #199

### DIFF
--- a/web/handlers/admin/invites.go
+++ b/web/handlers/admin/invites.go
@@ -23,8 +23,6 @@ type invitesHandler struct {
 
 	db     roomdb.InvitesService
 	config roomdb.RoomConfig
-
-	domainName string
 }
 
 func (h invitesHandler) overview(rw http.ResponseWriter, req *http.Request) (interface{}, error) {
@@ -73,8 +71,6 @@ func (h invitesHandler) create(w http.ResponseWriter, req *http.Request) (interf
 	}
 
 	facadeURL := h.urlTo(router.CompleteInviteFacade, "token", token)
-	facadeURL.Host = h.domainName
-	facadeURL.Scheme = "https"
 
 	return map[string]interface{}{
 		"FacadeURL": facadeURL.String(),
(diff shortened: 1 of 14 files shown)


web/handlers/admin/invites.go
```diff
@@ -73,8 +73,6 @@ func (h invitesHandler) create(w http.ResponseWriter, req *http.Request) (interf
 	}
 
 	facadeURL := h.urlTo(router.CompleteInviteFacade, "token", token)
-	facadeURL.Host = h.domainName
-	facadeURL.Scheme = "https"
```
staltz, line 77: What happened here? :)

web/templates/invite/facade.tmpl
```diff
@@ -1,7 +1,7 @@
 {{ define "title" }}{{ i18n "InviteFacadeTitle" }}{{ end }}
 {{ define "content" }}
 <div class="flex flex-col justify-center items-center self-center max-w-lg">
-  <p id="welcome" class="text-center mt-8 italic">Welcome to {{.RoomTitle}}! {{i18n "InviteFacadeWelcome"}}</p>
+  <p id="welcome" class="text-center mt-8 italic">{{i18nWithData "InviteFacadeWelcome" "RoomTitle" .RoomTitle}}</p>
```
staltz, line 4: :+1:

web/handlers/admin/invites.go
cryptix: `urlTo` was changed a while ago to set this itself. This here went un-noticd in the refactor and is obsolete now.

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/385b98a3a11a4316e1403e7c62ae7ec91a3f78aa/web/utils.go#L99-L104

staltz: Cool cool

staltz (approved):

Issue #110, opened 2021-03-29
Design an alternative to tunnel.endpoints


cryptix:
**Problem:**

muxrpc source `tunnel.endpoints` emits the _full_ list of connected peers for each change in connected peers. This is overhead intensive if people join and leave a lot. It sends n-1 (in case of leave, or n+1 in case of a join) peer identifiers for each connected peer.

Additionally, if an UI want's to do something _nice_ with the list of peers (like who joined or left last) it has to keep the state for comparison anyhow so the _reactive_ argument of not having to hold the state doesn't really count and shouldn't count for a network protocol in any case.

**Solution:**

Just emit state changes in the form of `{type: left, id: @their=.feed}` with valid values for `type`: `joined` and `left`.

(`type:` could be `state:` as well, don't really care about the specific name(s) that much. `joined` could be `entered`, etc.)

**Name** for this new rpc: `room.attendants`

updates #89

staltz:
Yes, I'm favorable of this, but I think there's something missing: how would the client know what is the initial state? Is there a separate API to call to get the initial state?

E.g. suppose the current state of attendants is `[A,B,C]` and then D joins, and then E joins. When D joins, they'll subscribe to `room.attendants` and they'll be notified `{type: 'joined', id: E}`, but it's missing A, B, C. Have you thought about this? I had some thoughts but haven't found a way that's resistant against race conditions (subscribing too slow to the updates).

staltz:
One idea: the first event emitted on `room.attendants` (`source` muxrpc I suppose) is the initial state, e.g. `{type: "state", [idA, idB, idC]}` and subsequent emissions are `type: joined` or `type: left`.

cryptix:
Oh, totally right! I didn't mention initial state at all... :see_no_evil: 

+1to your suggestion to emit it once in the beginning!

staltz:
Good, so here's how it would look like (and I might try implementing it in ssb-room-client already):

- Name: `room.attendants`
- Type: `source`
- Data: 
  1. Initial state `{"type":"state","ids":[A,B,C]}`
  2. Update
    - `{"type":"joined","id":D}`
    - `{"type":"left","id":B}`

cryptix:
I feel like we are in a good place here and will implement this now.

PR #218, opened 2021-05-17
add metadata to isRoom

### LINKED ISSUE #58: Room metadata when peer calls isRoom()
**Actual response:** `boolean`

**Expected response:**

```
{
  name: 'hermies.club', // i.e. the room's host
  features: [
    'tunnel', 
    'room1',
    // 'room2', // once we have attendants() and metadata()
    'alias',
    'httpAuth',
    'httpInvite'
  ]
}
```

-------

From Wouter's [UX research](https://www.manyver.se/ux-research/):

> Another note for a server feature: in the design of the Join room server screen, I conceptualised a new feature for room servers. Where a room server admin could link their personal SSB account to the server. This is expected to increase trust for potential new users of this room server, because they can then see which person owns it. (See the notes in report 3)

Actually, room1 answers `rpc.tunnel.isRoom` with an object which gives metadata about the room:

https://github.com/staltz/ssb-room/blob/e78d54887682664def36d48ca9e648fc609478e9/tunnel/server.js#L71-L77

For room2, we could support the same, while also adding more metadata, such as the admin's feed id, and the moderators feed ids.

cryptix:
Fixes #58

### DIFF
--- a/muxrpc/handlers/tunnel/server/state.go
+++ b/muxrpc/handlers/tunnel/server/state.go
@@ -5,6 +5,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -12,7 +13,6 @@ import (
 	"github.com/ssb-ngi-pointer/go-ssb-room/internal/network"
 	"github.com/ssb-ngi-pointer/go-ssb-room/roomdb"
 	"github.com/ssb-ngi-pointer/go-ssb-room/roomstate"
-	refs "go.mindeco.de/ssb-refs"
 
 	kitlog "github.com/go-kit/kit/log"
 	"go.cryptoscope.co/muxrpc/v2"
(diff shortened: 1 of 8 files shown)


muxrpc/handlers/tunnel/server/state.go
```diff
@@ -5,30 +5,74 @@ package server
...
+
+	var reply IsRoomReply
+	reply.Name = h.netInfo.Domain
+
+	// check if caller is a member
+	if _, err := h.members.GetByFeed(ctx, *ref); err != nil {
+		if errors.Is(err, roomdb.ErrNotFound) {
+			reply.Membership = false
+		}
+		return nil, err
```
staltz, line 55: Don't we need an `else` here?

cryptix: It was a bit convoluted, i think and hope the new version is clearer.

staltz: looks good!

cryptix:
@staltz if you have a JS module to check the format, we could add a test using that, too.

Elsewise this is good to go, i think.

staltz (approved):
IMO ready to go! Regarding tests, we could use ssb-room-client which already has some rudimentary tests for these metadata: https://github.com/ssb-ngi-pointer/ssb-room-client/blob/02b54773c78a9bb4fe77b933777bb1eb5a3bf33b/test/success-cases.js#L83-L88

Something like 

- go-ssb-room creates invite
- Node.js client accepts an invite from the Go server using ssb-http-invite-client
- Node.js client connects to the go-ssb-room
- isRoom happens in the background and Node.js client saves data in conn.json
- We check what was written to conn.json

But this can be done separately from this PR, if you're looking forward to press this green button now.

cryptix:
I think my prio no1 is to unblock you with the stuff you need for the community launch and so will take a spin on the go side of #110 next instead of this integration test.

Issue #89, opened 2021-03-17
Use new muxrpc namespace room.* and soft deprecate tunnel.*


staltz:
- [x] Design/specify `room.metadata`
  - Should list which *features* the room supports, e.g. #169 
- [x] Design/specify `rooms.attendants`, an alternative to `tunnel.endpoints`
  - See #110 
- [x] Implement `room.metadata`
  - See #58 
- [x] Implement `room.attendants` alternative, #110
- [x] Mark `tunnel.endpoints` as deprecated but keep it working
- [x] Mark `tunnel.isRoom` as deprecated but keep it working
- [x] remove `room.announce` and `room.leave`
- [x] Update the rooms2 spec doc

staltz:
> tunnel.endpoints alternative

Ideas:

- `room.crowd`
- `room.attendance`
- `room.attendants` (my favorite)
- `room.audience`
- `room.assemblage`

cryptix:
`room.attendants` :+1: (had to look it up but contrary to my thoughts it translates to the german term for the right concept)

staltz:
> Mark tunnel.isRoom as deprecated but keep it working
 
 is this done?

cryptix:
See https://github.com/ssb-ngi-pointer/go-ssb-room/pull/220#issuecomment-842401241.

I wasn't sure if those deprecation comments were meant for the JS side. There isn't really a defined way to do deprecation in Go. You usually just remove the APIS from the next major version.

staltz:
@cryptix does attendants include the client's ID too? It seems like it doesn't? That's fine if it's the case, it's just different from tunnel.endpoints and I'd need to know (also to update the spec).

cryptix:
Sorry, I'm not following. Is the client the joining/parting room member or the room or..?

staltz:
When Alice connects to the room and calls the `room.attendants()` stream, she gets `{type: "state", ids: [BOB, CARLA]}` not `{type: "state", ids: [ALICE, BOB, CARLA]}`.

staltz:
spec update: https://github.com/ssb-ngi-pointer/rooms2/pull/43

PR #223, opened 2021-05-21
remove room.announce and room.leave from muxrpc

cryptix:
Those two were confusion and implied some use beyond just opening `tunnel.endpoints`/`room.attendants` (which implicitly does what `tunnel.announce` does).

updates #89


staltz (approved):
Thanks!

PR #228, opened 2021-05-31
Small typo fix in the README link of hosting

HendrikPetertje:
The nginx-example conf has a different name in the repo


cryptix:
Thanks!

PR #239, opened 2021-06-01
add Dockerfile

cryptix:
as requested in #237

What's missing is documentation on how to use this with nginx, which I'm honestly not really well versed in.

Also: a better way to parametrize the TLS domain. I guess one would use `-e something=my-room.domain` ?

PR #240, opened 2021-06-01
Add docker and docker-compose with instructions for people that want to deploy that way.

HendrikPetertje:
Its a fairly simple setup

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.16-alpine
+
+RUN apk add --no-cache \
+      build-base \
+      git \
+      sqlite \
+      sqlite-dev
+
+RUN mkdir /app
+WORKDIR /app
+COPY . /app
+
+RUN cd /app/cmd/server && go build && \
+    cd /app/cmd/insert-user && go build
+
+EXPOSE 8008
+EXPOSE 3000
+
+RUN apk del \
+      build-base \
+      git
+
+CMD ./start.sh
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -21,6 +21,46 @@ After running `sudo dpkg -i go-ssb-room_v1.2.3_Linux_x86_64.deb` pay special att
 * You should now have a working go-ssb-room binary! Read the HTTP Hosting section below and admin
   user sections below, for more instructions on the last mile.
 
+# Docker & Docker-compose
+
+This project includes a docker-compose.yml file as well as a Docker file. Using
+it should be fairly straight forward.
+
+Start off by making a copy of `.env_example` called `.env` and insert your 
+website domain there. With that done execute
+
+```
+docker-compose build room
+```
+
+Followed by
+
+```
+docker-compose up
+```
+
+Your database, secrets and other things will be synchronized to a folder in your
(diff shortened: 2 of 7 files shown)


HendrikPetertje:
I could probably run a start.sh script that fetches the domain (and other settings) from the system environment and then inject those when people do `docker start` instead of docker-compose commands. That way the Dockerfile itself could have a CMD.

HendrikPetertje:
included that now

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+      git \
+      sqlite \
+      sqlite-dev
+
+RUN mkdir /app
+WORKDIR /app
+COPY . /app
+
+RUN cd /app/cmd/server && go build
+
```
ahdinosaur, line 15: do we want to:

```dockerfile
EXPOSE 8008
EXPOSE 3000
```

?

docs/deployment.md
```diff
@@ -21,6 +21,48 @@ After running `sudo dpkg -i go-ssb-room_v1.2.3_Linux_x86_64.deb` pay special att
...
+```
+
+Followed by
+
+```
+docker-compose up
+```
+
+Your database, secrets and other things will be synchronized to a folder in your
+project called "docker-secrets".
```
ahdinosaur, line 43: why `docker-secrets` and not `ssb-go-room-secrets`?

docs/deployment.md
```diff
@@ -21,6 +21,48 @@ After running `sudo dpkg -i go-ssb-room_v1.2.3_Linux_x86_64.deb` pay special att
...
+
+```
+docker-compose exec room sh
+```
+
+Then inside the virtual machine:
+
+```
+cd /app/cmd/insert-user
+go build
```
ahdinosaur, line 57: should we just build this in the Dockerfile or no? was also wondering why we keep all our build dependencies in the Dockerfile after build, but now i see we need them for this.

ahdinosaur (approved):
in general looks great, especially as a foundation to get started! :watermelon:

Dockerfile
HendrikPetertje: We should yes, good one!

docs/deployment.md
HendrikPetertje: I can go with ssb-go-room-secrets ;) I'll make some edits in a bit

docs/deployment.md
HendrikPetertje: jep, you mean that you'd want to remove go from the final build... that might not be the worst idea!

HendrikPetertje:
I have now added a bunch of small fixes as @ahdinosaur has pointed out. I choose not to remove go, it seems that the off-the-shelf alpine image does all kinds of things with GO that I didn't have time to retrace. Aditionally Luaandro had a genius idea of making it possible to include ARM support.

Before I dive into that however I would love to get this one merged (so we can build on top of it).

cryptix:
>  I chose not to remove GO,

I take it I'll close #239 then, too?

> Before I dive into that however I would love to get this one merged (so we can build on top of it).

understood.

(as stated previously, all of this is quite opaque to me but I welcome the contributions a lot!)

Issue #246, opened 2021-06-04
Document about necessity of allowing 8008/tcp


romuloalves:
It would be good to document in the deployment docs about the necessity of allowing port 8008 over TCP in order to enable `Sign-in with SSB` and connection to room as well.

PR #247, opened 2021-06-04
Add instruction about allowing TCP ports

romuloalves:
Adds instruction about allowing TCP ports because without them the room doesn't work.

Fixes https://github.com/ssb-ngi-pointer/go-ssb-room/issues/246.


staltz (approved):

cryptix:
Thanks!

Issue #250, opened 2021-06-09
Entries on /admin/members can overflow the UI layout due to multiple aliases


gtim:
On the Members page, having multiple aliases can make an entry overflow.

![image](https://user-images.githubusercontent.com/1546926/121424864-1a527b80-c972-11eb-82b4-f20bcee980d4.png)

I'd be happy to write a patch. Would you prefer truncating the alias list or expanding the entry?

staltz:
Thanks for spotting this! Yes we'd prefer the latter: expanding the height of the list item. But the list item needs a minimum height (matching the same fixed height dimensions as today).

PR #251, opened 2021-06-09
expand member entry when aliases overflow UI layout (#250)

gtim:
Should resolve issue #250. Side-by-side screenshots before and after changes on chrome:

image


staltz (approved):
Looks great, thanks!

@cryptix The CI is failing to build because `error: pathspec 'fix-multi-alias-overflow' did not match any file(s) known to git`, while the branch `fix-multi-alias-overflow` only exists on `gtim`'s repo, not here.

@gtim I see you updated `web/assets/style.css`, was that built for production? If so, we could ignore the red CI for now because it would anyway just build it for prod. Go tests are passing in CI.

gtim:
Thank you @staltz! I rebuilt style.css as Tailwind doesn't supply min-height-*x*. Thanks for checking, it was indeed built for development. In the last commit, it's built with `npm ci && npm run compile-prod`.

staltz:
Great! Now CI is green again :) (nice that a side effect is that besides building it for production, it can also *test* whether the PR is built for production)

PR #252, opened 2021-06-10
Exclude revoked invites in dashboard invite count (#177)

gtim:
This addresses #177. I'm not too familiar with Go traditions and SQLBoiler -- is the present approach with Invites.CountActive appropriate, or would adding a parameter to Invites.Count be preferred?

I'm yet to update the Invites mock and write a corresponding unit test.

Thanks a lot for your ssb work!

### DIFF
--- a/roomdb/sqlite/invites.go
+++ b/roomdb/sqlite/invites.go
@@ -250,18 +250,12 @@ func (i Invites) List(ctx context.Context) ([]roomdb.Invite, error) {
 	return invs, nil
 }
 
-func (i Invites) Count(ctx context.Context) (uint, error) {
-	count, err := models.Invites().Count(ctx, i.db)
-	if err != nil {
-		return 0, err
+func (i Invites) Count(ctx context.Context, onlyActive bool) (uint, error) {
+	queryMod := qm.Where("1")
+	if onlyActive {
+		queryMod = qm.Where("active = true")
 	}
-	return uint(count), nil
-}
-
-func (i Invites) CountActive(ctx context.Context) (uint, error) {
-	count, err := models.Invites(
-		qm.Where("active = true"),
-	).Count(ctx, i.db)
+	count, err := models.Invites(queryMod).Count(ctx, i.db)
 	if err != nil {
 		return 0, err
 	}
--- a/roomdb/sqlite/invites.go
+++ b/roomdb/sqlite/invites.go
@@ -258,6 +258,16 @@ func (i Invites) Count(ctx context.Context) (uint, error) {
 	return uint(count), nil
 }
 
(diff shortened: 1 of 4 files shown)


cryptix:
Cool! Thanks a bunch!

I think in this case I'd change the original `Count(...) ...` to something like `Count(_ context.Context, onlyActive bool) ...`

Having a separate function for it seems a bit too much. What would change this bargain (for me) is if `Count()` is used in a ton of places or had a bunch of arguments already. But it's pretty slim and only used for the dashboard anyhow.

gtim:
Thank you for sharing your thinking @cryptix! I've updated the branch with a single Count function as suggested, and updated the mock.

roomdb/sqlite/invites.go
```diff
@@ -250,8 +250,12 @@ func (i Invites) List(ctx context.Context) ([]roomdb.Invite, error) {
 	return invs, nil
 }
 
-func (i Invites) Count(ctx context.Context) (uint, error) {
-	count, err := models.Invites().Count(ctx, i.db)
+func (i Invites) Count(ctx context.Context, onlyActive bool) (uint, error) {
+	queryMod := qm.Where("1")
```
cryptix, line 254: You can also do this, then we don't need this default and if clause:

```go
qry := qm.Where("active = ?", onlyActive)
```

cryptix (commented):
One minor nitpick otherwise LGTM!

roomdb/sqlite/invites.go
gtim: Thanks a lot for your feedback! I read through the SQLBoiler docs to see if I could avoid that. If I understand correctly this suggestion would return the number of inactive invites for `onlyActive=false`, not the original behaviour of inactive+active?

cryptix: ah.. you are totally right. My suggestion was half-assed, indeed.

What _should_ work is this:

```go
	// have an empty query mods slice
	var queryMod []qm.QueryMod
	if onlyActive { // append to it if we want to filter
		queryMod = append(queryMod, qm.Where("active = true"))
	}
	// expand the slice to the variadic arg (works fine with zero qm's)
	count, err := models.Invites(queryMod...).Count(ctx, i.db)
	...
```

cryptix: but i'm honestly not sure I prefer that, feels clunky..

 having the tautological `true == true` (ie `1`) in there doesn't seem bad either

 merging this. thanks again!

gtim: Thanks you, I appreciate seeing your suggestions a lot!

Issue #254, opened 2021-06-24
AliasResolutionInstruct missing in active.de.toml


pathab:
If default language is German the alias page is empty and shouts an error: 

`alias-resolve render errr: render: template(alias.tmpl) execution failed: template: alias.tmpl:5:32: executing "content" at <i18n "AliasResolutionInstruct">: error calling i18n: i18n/error: failed to localize label AliasResolutionInstruct: message "AliasResolutionInstruct" not found in language "de"`

cryptix:
Sorry you ran into this and thanks for reporting.

I plan to spend a day on the localization feature next week and this will surely be on the list of things.

pathab:
Cool, thank you!

If you need more or better German translation, I can also help.

cryptix:
Yes, feel free!

I know some of them are a bit goofy. See also: https://github.com/ssb-ngi-pointer/go-ssb-room/issues/83#issuecomment-854355457

decentral1se:
Since https://github.com/ssb-ngi-pointer/go-ssb-room/pull/257#issuecomment-976465264 is merged, I think this can be closed?

PR #257, opened 2021-06-26
German translation

pathab:
Probably not perfect, but much better than before. I have already tested it on my server. There are still some texts that cannot be translated with the file (e.g. alias page). If new keys are added, I will be happy to update them. I'm looking forward to the further updates :-)


pathab:
#254 can be resolved with this translation-file.

cryptix:
> There are still some texts that cannot be translated with the file (e.g. alias page).

sorry, not sure i get this part..?

> If new keys are added, I will be happy to update them.

I don't think we plan to do so anytime soon. :)

otherwise looks great and I'd merge?

pathab:
image

Alias Page: Here I didn't find the corresponding key to translate.

staltz:
I'll merge based on cryptix's comment "otherwise looks great and I'd merge".

Thanks for your contribution!

Issue #135, opened 2021-04-09
Support Android *Chrome* custom URI links somehow


staltz:
The SSB URIs can't be clicked on Android Chrome, it just opens an empty page with the SSB URI filled in the address bar. Switching to Android Firefox, and everything works.

Unfortunately, *It Is NoT tHaT SiMpLe* and I'm not sure how to make it work. 

More info: 
- https://paul.kinlan.me/deep-app-linking-on-android-and-chrome/
- https://stackoverflow.com/questions/37069816/deep-linking-not-working-in-chrome
- https://stackoverflow.com/questions/49045919/why-is-url-with-custom-uri-scheme-not-launching-my-app-from-chrome-but-works-fin?noredirect=1&lq=1
- https://github.com/EddyVerbruggen/Custom-URL-scheme/issues/156

staltz:
For anyone who would like to help us fix this issue, here's a simple way of reproducing:

1. Make a webpage with an `<a>` link with the href equal to `ssb://experimental/?action=consume-alias&alias=staltz&multiserverAddress=net%3Ahermies.club%3A8008~shs%3AuMYDVPuEKftL4SzpRGVyQxLdyPkOiX7njit7%2BqT%2F7IQ%3D&roomId=%40uMYDVPuEKftL4SzpRGVyQxLdyPkOiX7njit7%2BqT%2F7IQ%3D.ed25519&signature=bP-z-DPIOZrWSVvHJdCfUZ2Zq30fyJ_rWLzIcknXi1Jg0aE1jPBVWofe3RnwxBRszYgWHPx7haCzByW89avLDQ%3D%3D&userId=%40QlCTpvY7p9ty2yOFrv1WU1AE88aoQc4Y7wYal7PFc%2Bw%3D.ed25519` or maybe another href that start with `intent://` (read this [stackoverflow](https://stackoverflow.com/questions/37069816/deep-linking-not-working-in-chrome))
2. Open that webpage on Android Chrome
3. Click the `<a>` link

To fix it, find a way, either with JS or with alternative `href`s, to make Android Chrome open Manyverse when clicking the link.

staltz:
As a first solution, this intent URI will only address `se.manyver`. In the future when there exists more than one SSB App on Android that can handle these URIs, we should render a list of buttons, one for each app. And only in the case of Android Chrome.

staltz:
[...]

PR #271, opened 2021-11-09
fix support for SSB URIs on Android Chrome

staltz:
Finally found a fix for #135 . TL;DR is that on Android Chrome we have to do `intent://PATH?QUERYPARAMS#Intent;scheme=ssb;end;` instead of `ssb://PATH?QUERYPARAMS`. See [Android Chrome docs](https://developer.chrome.com/docs/multidevice/android/intents/). So we discover the user agent and trigger this fix only for Android Chrome.

:heavy_check_mark: Unit tests added.

:heavy_check_mark: Tested end-to-end with a phone running Manyverse (tested both Android Firefox and Android Chrome).

### DIFF
--- a/web/handlers/aliases_test.go
+++ b/web/handlers/aliases_test.go
@@ -158,5 +158,5 @@ func TestAliasResolveOnAndroidChrome(t *testing.T) {
 	a.Equal(ts.NetworkInfo.MultiserverAddress(), params.Get("multiserverAddress"))
 
 	frag := aliasURI.Fragment
-	a.Equal("Intent;scheme=ssb;package=se.manyver;end;", frag)
+	a.Equal("Intent;scheme=ssb;end;", frag)
 }
--- a/web/handlers/aliases_test.go
+++ b/web/handlers/aliases_test.go
@@ -100,3 +100,63 @@ func TestAliasResolve(t *testing.T) {
 	html, resp = ts.Client.GetHTML(htmlURL)
 	a.Equal(http.StatusInternalServerError, resp.Code)
 }
+
+func TestAliasResolveOnAndroidChrome(t *testing.T) {
+	ts := setup(t)
+
+	a := assert.New(t)
+	r := require.New(t)
+
+	var testAlias = roomdb.Alias{
+		ID:   54321,
+		Name: "test-name",
+		Feed: refs.FeedRef{
+			ID:   bytes.Repeat([]byte{'F'}, 32),
+			Algo: "test",
+		},
+		Signature: bytes.Repeat([]byte{'S'}, 32),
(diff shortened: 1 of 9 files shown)


staltz:
@cryptix @cblgh Could one of you take a look at this PR? (i.e. OR is what I'll mean when I tag you both as reviewers, not AND and not XOR :smiley_cat:)

cblgh:
@staltz i was intending to get to this today but alas the gods would not have it so. will give it a review next week unless @cryptix beats me to it!

web/handlers/aliases_test.go
```diff
@@ -100,3 +100,63 @@ func TestAliasResolve(t *testing.T) {
...
+			Algo: "test",
+		},
+		Signature: bytes.Repeat([]byte{'S'}, 32),
+	}
+	ts.AliasesDB.ResolveReturns(testAlias, nil)
+
+	// Mimic Android Chrome
+	var uaHeader = make(http.Header)
+	uaHeader.Set("User-Agent", "Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5 Build/MOB30H) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.133 Mobile Safari/537.36")
+	ts.Client.SetHeaders(uaHeader)
```
cryptix, line 124: 👌🏻

web/handlers/invites.go
```diff
@@ -39,19 +40,31 @@ type inviteHandler struct {
...
+	// https://github.com/ssb-ngi-pointer/go-ssb-room/issues/135
+	browser := ua.Parse(userAgent)
+	if browser.IsAndroid() && browser.IsChrome() {
+		joinRoomURI = url.URL{
+			Scheme:   "intent",
+			Opaque:   "//experimental",
+			RawQuery: queryVals.Encode(),
+			Fragment: "Intent;scheme=ssb;end;",
+		}
+	}
```
cryptix, line 67: I'd prefer to have this and the one in aliases in one place but I wouldn't call veto for it.

`web/utils.go` is a bit of a grab bag of things right now. Maybe it could be a function, there?

cryptix (commented):
LGTM

just one nit-pick.


super glad and excited you found a solution for this (and also, wtf chrome?)!

web/handlers/invites.go
staltz: Thanks! I also thought it should be centralized somewhere but was waiting for a good suggestion. I think I'll do that

staltz:
Ready for a second check!

staltz:
I'll merge assuming that the review was green apart from that refactor.

Issue #272, opened 2021-11-11
Can't change listening port with server parameters


brunovianna:
To replicate: run server with -lishttp with a different port then 3000

` ./server  -lishttp :5555`

also tried 

` ./server  -lishttp 5555`

The server continued to listen on port 3000. To double check, I changed the source code line to 5555, complied and ran again, and it listened on port 5555

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/794a8f9755991094c2986e177a8b21f564176e30/cmd/server/main.go#L98

I'm trying to debug further to propose a PR.

brunovianna:
Quick update: it is actually the last parameter that is not parsed, not necessarily lishttp

brunovianna:
ok, the problem is that the format 
`
-flag x`

only works for non-booleans. since -aliases-as-subdomains is a boolean and is used in 
https://github.com/ssb-ngi-pointer/go-ssb-room/blob/794a8f9755991094c2986e177a8b21f564176e30/start.sh

as 
` -aliases-as-subdomains "${ALIASES_AS_SUBDOMAINS}"`

it didn't work and was messing up other args. I just made a PR with small fixes.

PR #273, opened 2021-11-11
proposed fix for issue #272. args were not correctly formatted. 

brunovianna:
Just changed the args formatting in start,sh
And also change a reference that was made in the development documentation (although it was correctly formatted, it could lead to mistakes)


staltz:
Great! Thank you! :)

PR #266, opened 2021-10-08
Deps and release fix

cryptix:
fixes #241

Issue #241, opened 2021-06-01
include insert-user tool in releases


cryptix:

mhfowler:
just ran into this while trying to package go-ssb-room for yunohost. as I understand it, the deb or the tar.gz in the release won't currently be enough to get things going without insert-user. 

on yunohost forum, they also like if builds for multiple architectures are included (here is [a thread about that](https://forum.yunohost.org/t/recommended-way-to-package-golang-app/17092))

I'm not sure what tooling you are currently using to make the release, but maybe we could automate it so that it makes the builds for multiple architectures and includes insert-user in all of them. I'd be happy to work on that if you can tell me what tools you use for releases @cryptix

decentral1se:
https://github.com/ssbc/go-ssb-room/pull/355#issuecomment-1312091627

Issue #79, opened 2021-03-12
github.com/mattn/go-sqlite3 requires CGo


cryptix:
> cryptix: if the cgo based sqlite ever becomes a problem, there always is madscience: https://gitlab.com/cznic/sqlite
cryptix: o/ is a machine translated version of the C "dialect" in sqlite to pure Go. so it can be cross-compiled and used in environments where cgo is difficult or unavailable. possibly even wasm
statlz: if in wasm, would it run in the browser, but ... where would it store the data?
cryptx: not saying we need a wasm room but i disregarded sqlite before for other projects becasue cgo can be a pain
cryptix: like, if i wanted go-ssb-room on my small FreeBSD NAS, I need to compile for FreeBSD with a full C environment.. installing Go on the box isn't even enough then but also a full gcc/clang stdenv

The sqlite driver is the only CGo dependency we have, everything else can be cross-compiled.

I'm not jumping ship for the new, cgo-free version because I'm quite sure it still has some teething problems to go through but it's super impressive to see.

related links:

* https://dave.cheney.net/2016/01/18/cgo-is-not-go
* https://blog.golang.org/cgo
* https://gitlab.com/cznic/sqlite
* https://github.com/mattn/go-sqlite3

ahdinosaur:
would folx be keen for arm64 cross-compilation? i'm keen to run butt.nz` on a Raspberry Pi. :pie:

decentral1se:
Gitea are currently going for it in https://github.com/go-gitea/gitea/pull/20614 so I think it might be worth a look again!

PR #291, opened 2022-01-10
Cross-compile arm releases & include insert-user in release

ahdinosaur:
hi, i'm keen for `go-ssb-room` on a Raspberry Pi.  :pie: 

so i improved the release dev tools, using [Vagrant](https://www.vagrantup.com/) (a friendly wrapper around VirtualBox).

open to feedback! :purple_heart: 

related:
-  https://github.com/ssb-ngi-pointer/go-ssb-room/issues/79

also now i see https://github.com/ssb-ngi-pointer/go-ssb-room/pull/266 and https://github.com/ssb-ngi-pointer/go-ssb-room/issues/241, happy to incorporate other changes as need be, could also cross-compile `armel` and `armhf` if need be.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -29,7 +29,7 @@ builds:
       - linux
     goarch:
       - amd64
-  - id: go-ssb-room-insert-arm64
+  - id: go-ssb-room-linux-arm64
     env:
       # needed for sqlite
       - CGO_ENABLED=1
@@ -55,13 +55,13 @@ builds:
       - linux
     goarch:
       - arm64
-  - id: go-ssb-room-insert-armel
+  - id: go-ssb-room-linux-armhf
     env:
       # needed for sqlite
       - CGO_ENABLED=1
       # cross-compilation
-      - CC=arm-linux-gnueabi-gcc
-      - CXX=arm-linux-gnueabi--g++
+      - CC=arm-linux-gnueabihf-gcc
+      - CXX=arm-linux-gnueabihf--g++
     main: ./cmd/server
     binary: go-ssb-room
     goos:
@@ -71,13 +71,13 @@ builds:
     goarm:
       - 6
       - 7
-  - id: go-ssb-room-insert-user-linux-armel
+  - id: go-ssb-room-insert-user-linux-armhf
     env:
       # needed for sqlite
       - CGO_ENABLED=1
       # cross-compilation
-      - CC=arm-linux-gnueabi-gcc
-      - CXX=arm-linux-gnueabi--g++
+      - CC=arm-linux-gnueabihf-gcc
+      - CXX=arm-linux-gnueabihf--g++
     main: ./cmd/insert-user
     binary: go-ssb-room-insert-user
     goos:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -9,7 +9,7 @@ before:
     - go generate ./...
 
 builds:
-  - id: linux-amd64
+  - id: go-ssb-room-linux-amd64
     env:
       # needed for sqlite
       - CGO_ENABLED=1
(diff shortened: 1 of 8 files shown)


.goreleaser.yml
```diff
@@ -31,9 +45,6 @@ gomod:
 nfpms:
  - id: go-ssb-room
    package_name: go-ssb-room
-   replacements:
```
ahdinosaur, line 34: i removed these replacements because they seem unnecessary, i'd argue the default names are more clear because they match everything else. but happy to revert this.

.goreleaser.yml
```diff
@@ -44,9 +55,10 @@ nfpms:
    suggests:
      - ngingx
      - certbot
-   empty_folders:
-     - /var/log/go-ssb-room
    contents:
+     - dst: /var/log/go-ssb-room
```
ahdinosaur, line 59: `goreleaser` complained that `empty_folders` was deprecated, so did this

ahdinosaur:
this is ready for review! :cake: 

here's some compiled releases: https://github.com/ahdinosaur/go-ssb-room/releases/tag/v2.1.0-pre.1

/cc @mhfowler

docs/files/debian-postinstall.sh
```diff
@@ -1,11 +1,43 @@
...
 #
 # SPDX-License-Identifier: CC0-1.0
 
+set -e
+
 # create a user to run the server as
 adduser --system --home /var/lib/go-ssb-room go-ssb-room
 chown go-ssb-room /var/lib/go-ssb-room
 
+if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
```
ahdinosaur, line 13: i stole these systemd script things from [peachcloud/peach-package-builder](https://github.com/peachcloud/peach-package-builder) who stole them from [mmstick/cargo-deb](https://github.com/mmstick/cargo-deb) :package:

ahdinosaur: however, they do change things such that the service starts automatically when installed (and is enabled to start on boot), as this is the usual behavior on Debian. if we follow this behavior, we should update the postinstall message to tell you to restart the service rather than reload daemon after making changes.

would there be a problem with the service first running with the wrong config, then changing it and restarting the service?

as a side note, we shouldn't be recommending updating services by writing to the systemd file directly, there's a whole built-in edit workflow with conf overrides and such that we should be doing.

decentral1se (commented):
Nice work! I'm curious if it is possible to avoid the vagrant/vm requirement and instead use Docker? As far as I understand, you're using a Debian image & running a series of build commands which could be done with Docker?

ahdinosaur:
> I'm curious if it is possible to avoid the vagrant/vm requirement and instead use Docker? As far as I understand, you're using a Debian image & running a series of build commands which could be done with Docker?

:shrug: 

i usually use Docker, but here i used Vagrant and had a lovely time. :smiley_cat:

.goreleaser.yml
decentral1se: Seems legit, let's use the defaults.

decentral1se:
@ahdinosaur 

Thanks for this PR. I'd love to get it out in the near future and am doing a review now.

> i usually use Docker, but here i used Vagrant

Would you be up for using Docker instead? I think I'd love to avoid having to work with Vagrant going forward because it assumes a workstation that has virtualisation wired up and sometimes that is a bit tricky / can be a barrier to working on this. We could wire up a Debian-based Dockerfile and acommpanying commands instead? I could do that or if you feel like it, please go for it.

docs/files/debian-postinstall.sh
decentral1se: These changes seem legit! Thanks for adding.

> would there be a problem with the service first running with the wrong config, then changing it and restarting the service?

According to:

https://github.com/ssbc/go-ssb-room/blob/641069a8f176c5d3492b37a8c5ca6b46d1acd3fd/docs/files/debian-postinstall.sh#L26

The startup will fail? Maybe we can avoid automatically starting up & recommend the commands? Unsure what Debian does when installed packages need to be configured first?

> as a side note, we shouldn't be recommending updating services by writing to the systemd file directly, there's a whole built-in edit workflow with conf overrides and such that we should be doing: https://askubuntu.com/questions/659267/how-do-i-override-or-configure-systemd-services.

One for the docs I guess, can do here or let's open an issue to catch that another time?

ahdinosaur:
@decentral1se great, i'm happy for you to continue from where i left off, i'll be here on the sidelines if you need me. 😺

decentral1se:
Cut a trimmed version of this in https://github.com/ssbc/go-ssb-room/pull/355. Opened https://github.com/ssbc/go-ssb-room/issues/354 also.

PR #293, opened 2022-01-16
fix missing domain when subdomains are not used

auge:
when not using subdomains (-aliases-as-subdomains false), the generated link/html in the members list is "https:///alias/xxx" missing the domain string.
this small fix adds the domain string in front of the generated URL.
hope this is the intended way (I'm not really a go programmer…)

KR, Benjamin

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -45,7 +45,7 @@ import (
 
 // Version and Build are set by ldflags
 var (
-	version = "v2.0.7-snapshot"
+	version = "v2.0.7"
 	commit  = "unset"
 )
 


staltz:
Thanks for the PR! The diagnostics sounds about right, and the fix is mostly correct, although a test would make it really good.

internal/network/interface.go
```diff
@@ -55,7 +55,7 @@ func (sed ServerEndpointDetails) URLForAlias(a string) string {
 	if sed.UseSubdomainForAliases {
 		u.Host = a + "." + sed.Domain
 	} else {
-		u.Path = "/alias/" + a
+		u.Path = sed.Domain + "/alias/" + a
```
staltz, line 58:
```suggestion
		u.Host = sed.Domain
 		u.Path = "/alias/" + a
```

I think it would be safer to do this instead.

auge: thanks for the review. I see and understand your comments.
I will be able to work on a test probably by tomorrow.

cryptix:
thanks for catching this and sending in the PR @auge!

> a test would make it really good

just FYI, it might be a bit hard to see where this would go. You would add a new Test.. func to `web/handlers/aliases_test.go` and change the the session that `setup(t)` returns to suite your needs (unset the domain, do what the flag does, etc).

PR #294, opened 2022-01-17
Fix alias generation

auge:
thanks for the feedback from PR #293.
I've now created a test function which fails without the following commit showing also the problem.
the even simpler fix as proposed by @staltz is included in the successive commit.

KR, Benjamin


auge:
I had to add a new member to the testSession struct to store the ServerEndpointDetails (basically copying over from web/handlers/admin/members_test.go)
new is the function TestURLForAlias, which first tests a subdomain alias and then the path-based alias

staltz:
Perfect! Provided that tests pass in CI, I'm in favor of merging this. Thank you @auge !

Issue #299, opened 2022-03-07
Accepting an invite when you're already a member


staltz:
**Expected:** the invite is accepted (claimed and removed) and the usual responses are given to the client.

**Actual:** the invite is rejected. (Tested with Manyverse 0.2203.7)

cc @achou11

cryptix:
This should be fairly simple patch, handling the error here:

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/6f5edbaa438fe473cc865e283bb3fd32de231c44/roomdb/sqlite/invites.go#L132-L135

Just make sure it actually is the right error type:

https://github.com/ssb-ngi-pointer/go-ssb-room/blob/7b3a5852dafa98b071e26ffaea9ae453db222c11/roomdb/sqlite/members.go#L74

So basically changing the error above to something like this:

```go
var alreadyAdded roomdb.ErrAlreadyAdded
if errors.As(err, &alreadyAdded) && alreadyAdded.PubKey.Equal(newMember) {
  // it is fine to use an invite twice
} else {
  return err
}
```

plus adding a test of course.. 🙃

PR #302, opened 2022-03-26
fix: accept invites as member is ok

decentral1se:
Took a stab at fixing https://github.com/ssb-ngi-pointer/go-ssb-room/issues/299.


staltz:
Ah, the smell of a perfect PR for something I've been procrastinating on... Thank you!

(I'm on mobile now, will check on my computer later)

staltz:
:tada:

Issue #275, opened 2021-11-13
Error: "authentication credentials incorrect", when they should be correct


staltz:
> Huge fan of the project!
>
> Got a room set up, made the first admin user, went to login with password, and got this error:
>
>  * The supplied authentication credentials (username or password) are incorrect.
>
> Tried adding the user again with a simpler password (thought maybe I was entering it incorrectly the first time around), but still getting the same error.
>
> I've attached the logs, not really sure what else to try.
>
> Thanks for your help!

![ssb_room_error_logs](https://user-images.githubusercontent.com/90512/141643968-44ae6ad5-5385-4df1-bf78-ac04eaed6833.PNG)

Count042:
I'm having the same issue, having recently deployed a room for an organization I run.

Is there a way I can get a room up and running if I can't login to it as an administrator to configure?  Is there a known working version I can deploy in the meantime?

Thanks!

geoah:
Just a hint in case it helps anyone who finds this issue: Make sure you are using the correct `-repo` path. The `insert-user` tool will not complain if you give it the wrong path. If you're using the docker-compose instructions, make sure you have uncommented the `REPO=` line in your `.env` file as it assumes you did so.

decentral1se:
@geoah do you think we could make `insert-user` more robust if it 1. made sure the `-repo` directory existed and/or did some validation 2. ensured that the `REPO=...` env var is present in the env? That might catch some of these issues & are small & easy fixes.

decentral1se:
Also worth noting in case it is not clear, you'll want to use your pubkey as username:

![image](https://user-images.githubusercontent.com/1991377/161870219-9b397d36-1592-4a90-9f4a-2ef671c36f06.png)

![image](https://user-images.githubusercontent.com/1991377/161870261-97117d55-76db-4991-a3fa-f35f5f309101.png)

Seems to me like we might want to rename "username" in the UI to be "public key"?

staltz:
> Seems to me like we might want to rename "username" in the UI to be "public key"?

Agreed!

PR #304, opened 2022-04-05
docs: wording

decentral1se:
Tiny one while looking into https://github.com/ssb-ngi-pointer/go-ssb-room/issues/275.


staltz:
Thank you!

PR #308, opened 2022-04-06
fix: error out if repoPath doesn't exist; user-facing username/public key -> SSB-ID

decentral1se:
This might close https://github.com/ssb-ngi-pointer/go-ssb-room/issues/275?

(My german is awful, sorry :see_no_evil:)

image

image

### DIFF
--- a/cmd/insert-user/main.go
+++ b/cmd/insert-user/main.go
@@ -53,6 +53,13 @@ func main() {
 	})
 	flag.Parse()
 
+	if _, err := os.Stat(repoPath); err != nil {
+		if os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "error: %s does not exist (-repo)?\n", repoPath)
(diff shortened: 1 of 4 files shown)


decentral1se (commented):
Although hmmm wait a sec, I see in https://github.com/ssb-ngi-pointer/go-ssb-room/blob/master/web/i18n/defaults/active.en.toml that "public key", "SSB identity" and "SSB ID" are used. Do we want to only use one? I might be missing the difference but I think they're all the same or?

staltz:
They're all the same, yeah. SSB ID would be my default choice, but I don't mind the others.

decentral1se:
OK maybe https://github.com/ssb-ngi-pointer/go-ssb-room/pull/308/commits/6c93b004bf771b701e60793a56e9ccb3c3cc98d6 works then, amended the commit.

PR #310, opened 2022-04-09
fix: username -> SSB-ID

decentral1se:
Small fix-up after https://github.com/ssb-ngi-pointer/go-ssb-room/pull/308. If you enter your deets wrong it tells you that might have got your username wrong but since the form now says SSB Identifier, it can be confusing. This resolves that confusion, I hope.


decentral1se:
I don't quite understand how the tests work and can't understand why this is failing :thinking: :see_no_evil:

decentral1se:
Bump, anyone got cycles to review / merge. It is a small one :crossed_fingers:

Issue #326, opened 2022-10-21
Documentation - path is wrong for insert-user when installed from .deb


KyleMaas:
When go-ssb-room is installed from the .deb file, the sample command for insert-user should be:

```
sudo ./insert-user -repo "/var/lib/go-ssb-room" "@Bp5Z5TQKv6E/Y+QZn/3LiDWMPi63EP8MHsXZ4tiIb2w=.ed25519"
```

decentral1se:
Yeh it'd be great to show the differences in a dedicated section for the .deb install alongside https://github.com/ssbc/go-ssb-room/blob/master/docs/deployment.md#docker--docker-compose :+1:

KyleMaas:
Fixed with #331

PR #331, opened 2022-10-21
Couple of documentation changes for Debian package

### LINKED ISSUE #324: Documentation - manually generate dhparams
Following through the installation docs (which are quite good, I might add) on an Ubuntu 22.04.1 server, there's one issue I ran into that I think might be worth mentioning.  certbot did not generate a dhparams file.  Maybe add an example command for generating that?  Something like:

```openssl dhparam -out /etc/letsencrypt/ssl-dhparams.pem 2048```

KyleMaas:
Adds some clarifications for how to get started with go-ssb-room when installed from the Debian package.

Fixes #324 and #326

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -113,6 +113,13 @@ When the process is complete with `certbot`, pay attention to where the certific
 in the filesystem. If it's at `/etc/letsencrypt/live/hermies.club`, it's correct, otherwise you may
 need to rename it e.g. `hermies.club-0001` to `hermies.club`. 
 
+The example nginx configuration uses prebuilt Diffie-Hellman parameters.  You can generate these
+with the following command:
+
+```
+openssl dhparam -out /etc/letsencrypt/ssl-dhparams.pem 2048
+```
+
 Then restart your server, e.g. `systemctl restart nginx`.
 
 If you see such errors as the following when setting up your deployment:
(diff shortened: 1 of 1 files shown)


decentral1se (approved):

macornwell:
I am finding that this insert-user command does not exist on a debian install. The file is nowhere to be found. (this is blocking me from getting this server up).

KyleMaas:
Yeah, you have to follow the instructions under "First Admin user".  I added some supplementary directions in this change to make it more obvious how to follow those directions:

https://github.com/ssbc/go-ssb-room/pull/331/commits/c9adfee96fbd314b4c37c9062e0228a699c61ce6

macornwell:
So you're saying I need to pull down the source code to build the utility? Gotcha. This should be made clearer. Normally if a person isn't building from source they aren't going to have the source code. (The advantage of a deb). Would certaintly be better if this utility was baked in via the deb.... (another time).

The other problem is that this cannot currently be built. (I know this isn't your problem). Just an FYI.

`build github.com/ssb-ngi-pointer/go-ssb-room/v2/cmd/insert-user: cannot load embed: malformed module path "embed": missing dot in first path element`

KyleMaas:
It built for me.  But I agree - it should be included in the Debian package.  I ran into the same problem here:

https://github.com/ssbc/go-ssb-room/issues/327

Which ended up being a duplicate of:

https://github.com/ssbc/go-ssb-room/issues/241

This pull request is just about trying to help folks avoid some of the pitfalls I hit when I tried to install it.

KyleMaas:
Thanks!

PR #333, opened 2022-10-25
Setup vue frontend

chereseeriepa:
- sets the router to point to an index.html
- added files from `rooms-frontend`'s production build into web/assets folder
- updated some of the `CSP` config to allow the external graphql API and blobs


decentral1se:
:thinking:

PR #351, opened 2022-11-08
Add a way to list notices as JSON

boreq:
Previously notices could only be displayed as HTML. This commit makes it possible to request a list of notices as JSON. This can be used to programmatically display a description of a room server in SSB clients.

The behaviour is governed by a query parameter. To list notices as JSON set a query parameter "encoding" to "JSON" when listing notices (for example https://example.com/notice/list?encoding=json). This parameter was chosen instead of using the "Accept" header as similar behaviour is already exhibited by other endpoints (namely the invite mechanism).

### DIFF
--- a/web/handlers/http.go
+++ b/web/handlers/http.go
@@ -329,12 +329,13 @@ func New(
 
 	// notices (the mini-CMS)
 	var nh = noticeHandler{
+		render:  r,
 		flashes: flashHelper,
 
 		notices: dbs.Notices,
 		pinned:  dbs.PinnedNotices,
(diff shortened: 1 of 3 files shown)


decentral1se (approved):
LGTM! Now that I look at this code, it's a bit strange to see no `return` statements connected with the top-level functions doing rendering but that's just how it's done :upside_down_face:  Do appreciate that your approach follows the existing approach of other end-points, nice one. Will merge this tomorrow unless somebody else does / objections appear. Thanks!

PR #368, opened 2022-12-13
Add a JSON endpoint for generating invites

boreq:
When running in open mode invites can be freely generated by accessing /create-invite. This displays an HTML page which creates and displays an invite to the user.

This commit adds an additional way of creating invites in open mode. A POST request can be sent to the same /create-invite endpoint with the Accept header set to application/json. This returns a JSON response which contains an invite url.

The purpose of this change is to make automatic invite generation easier in SSB clients.


decentral1se (approved):
LGTM! Will merge tomorrow if nobody else has concerns, thanks!

staltz (approved):

PR #387, opened 2023-06-25
change mattn/go-sqlite3 to modernc.org/sqlite

### LINKED ISSUE #79: github.com/mattn/go-sqlite3 requires CGo

cryptix:
As far as I can tell the modernc.org/sqlite transpile project came quiet a bit further, stabilised and proofed that it's much less mad science than people, me included, initially thought.

fixes #79

I will put this on hermies.club in a bit (if WifionICE allows me to). It's my understanding that this should be byte-per-byte compatible with the previous sqlite adapter and thus not need migration(s) but exhaustive testing is always in order with these things.


cblgh:
@cryptix oh sick!  does everything mostly just work after applying this patch? :o personal opinion: i think it might be wiser to do a merge & deploy after coming home from yr bike adventure :~

staltz:
Sounds okay by me! I don't have an opinion on this. Just try to make CI green before merging.

PR #395, opened 2023-07-16
made docker installation instructions more usable

DasMoorhuhn:
made some things more clear in the docker part :D


DasMoorhuhn:
By the way, the version from the golang image in the Dockerfile should be increased to version 1.18 because of the error, shown in https://github.com/ssbc/go-ssb-room/issues/394

